PR #133, opened 2026-05-21
fix(backend): force Chat Completions API for custom OpenAI-compatible endpoints

MesoX:
## Summary

- `sdk(id)` in `@ai-sdk/openai` v3+ resolves to the Responses API (`/v1/responses`), which uses a completely different request format (an `input` array containing `item_reference` and `function_call_output` objects)
- vLLM and other OpenAI-compatible servers only implement `/v1/chat/completions` and return a Python `KeyError('role')` 500 when receiving Responses API requests
- `sdk.chat(id)` explicitly targets `/v1/chat/completions`, which all providers handle correctly

## Test plan

- [x] Configure an OpenAI-compatible provider with a custom `baseUrl` (e.g. vLLM)
- [x] Send a message that triggers tool calls (e.g. an agent with MCP tools)
- [x] Confirm tool results are processed and the agent returns a final answer without a `'role'` error

### DIFF
--- a/apps/backend/src/services/provider.ts
+++ b/apps/backend/src/services/provider.ts
@@ -23,7 +23,7 @@ export const openProvider = (provider: Provider): OpenedProvider => {
         project: provider.project ?? undefined,
       });
       return {
-        languageModel: (id) => sdk.chat(id),
+        languageModel: (id) => (provider.baseUrl ? sdk.chat(id) : sdk(id)),
         embeddingModel: (id) => sdk.embeddingModel(id),
         searchTools: () => ({
           web_search: sdk.tools.webSearch({
--- a/apps/backend/src/services/provider.ts
+++ b/apps/backend/src/services/provider.ts
@@ -23,7 +23,7 @@ export const openProvider = (provider: Provider): OpenedProvider => {
         project: provider.project ?? undefined,
       });
       return {
-        languageModel: (id) => sdk(id),
+        languageModel: (id) => sdk.chat(id),
         embeddingModel: (id) => sdk.embeddingModel(id),
         searchTools: () => ({
           web_search: sdk.tools.webSearch({


willdady (changes requested):
> *This was generated by AI during review.*

Thanks for tracking this down — the diagnosis is spot-on and `sdk.chat(id)` is the right escape hatch.

One concern before merging: the PR title says "for custom OpenAI-compatible endpoints" but the diff is unconditional, so connections to the real OpenAI API also get switched off the Responses API. That means OpenAI-proper users silently lose Responses-only features — `previous_response_id` state, reasoning summaries from o-series models, and most concretely the hosted `web_search` tool that's still wired up at `provider.ts:29` via `sdk.tools.webSearch(...)`. Pairing a Responses-API hosted tool with a chat-completions model is likely either a no-op or a runtime error the next time someone enables web search on an OpenAI provider.

Could we gate on `baseUrl`? Something like:

```ts
languageModel: (id) => (provider.baseUrl ? sdk.chat(id) : sdk(id)),
```

That matches what the title and description claim, fixes vLLM and other OpenAI-compatible servers, and avoids regressing the OpenAI-proper path (including `web_search`).

Happy to merge once that's in — or if there's a reason you'd prefer the unconditional switch, let me know and we can discuss.

MesoX:
Hi,

yes, sorry, I thought OpenAI does have the same API as the local ones built on top of it, but seems it still uses the original one. I have commited a fix.  Thanks!

willdady:
Superseded by #140. After reflecting on the change-request discussion, the `baseUrl` heuristic is ambiguous — a non-null `baseUrl` doesn't reliably mean "not OpenAI" (proxies, gateways, LiteLLM, etc. all set custom base URLs while still talking to OpenAI-proper). #140 adds an explicit `apiMode` enum on the provider record instead, defaulting to `responses` so existing users see no behavior change. Thanks for the original diagnosis — the underlying `sdk.chat(id)` insight was correct and is preserved in the supersession.

PR #139, opened 2026-05-23
fix(backend): append access_type=offline for Google OAuth authorization URLs

MesoX:
## Summary

- Google only returns a refresh token when `access_type=offline` is present in the authorization URL — the `@ai-sdk/mcp` library never adds this parameter
- `prompt=consent` is added alongside it to force re-consent, which Google requires when granting offline access for the first time (even for previously-authorized accounts)
- The parameters are only injected for Google domains (`*.google.com`), so non-Google OAuth providers are unaffected

Without this fix, tokens expire after ~1 hour and the user must manually re-authorize in the MCP settings with no automatic recovery.

## Test plan

- [ ] Authorize a Google Drive or Google Calendar MCP via the OAuth flow
- [ ] Confirm `oauth_refresh_token` is stored in the `mcp` table after authorization
- [ ] Wait for the access token to expire (or manually clear it) and confirm the MCP continues working via automatic token refresh

### DIFF
--- a/apps/backend/src/services/mcp-oauth-provider.ts
+++ b/apps/backend/src/services/mcp-oauth-provider.ts
@@ -190,7 +190,13 @@ export class DatabaseOAuthClientProvider implements OAuthClientProvider {
   }
 
   async redirectToAuthorization(authorizationUrl: URL) {
-    // Capture the URL so the endpoint can read it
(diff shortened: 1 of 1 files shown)


willdady (approved):
LGTM. Targeted fix with correct hostname guard. Consider adding a unit test in a follow-up.

PR #144, opened 2026-05-27
fix(mcp): one-click Reauthorize via force flag, surface silent refresh as success

MesoX:
## Summary

The **Reauthorize** button on an MCP row reports `Failed to start OAuth authorization` whenever Platypus still holds a valid refresh token. The OAuth callback completes silently in the background but the UI treats it as a failure.

## Root cause

`POST /mcps/:mcpId/oauth/authorize` calls `mcpAuth(provider)` from `@ai-sdk/mcp`. When the stored refresh token is still valid the SDK silently rotates and returns `AUTHORIZED`; the handler then responds `{ message: "Already authorized" }` with no `authorizationUrl`. The frontend (`apps/frontend/components/mcp-form.tsx`) checks only for `data.authorizationUrl` and falls through to `toast.error(...)`. A successful silent refresh therefore looks identical to an error.

The user has no way to force a fresh OAuth flow short of editing the database to wipe tokens — `/oauth/revoke` exists but requires a separate click and confuses the intent.

## Fix

Two complementary changes:

1. **Backend** — add an opt-in `?force=true` query parameter to `POST /mcps/:mcpId/oauth/authorize`. When present, the handler clears `oauthAccessToken`, `oauthRefreshToken`, `oauthTokenExpiresAt`, `oauthScope` before invoking `mcpAuth`, so the flow always returns `REDIRECT` and yields an `authorizationUrl`. The DCR / static `oauthClientId` and `oauthClientSecret` are deliberately preserved so the same OAuth client is reused.

2. **Backend + frontend** — rename the silent-refresh response payload from `{ message: "Already authorized" }` to `{ alreadyAuthorized: true }`, and have the frontend treat that as success (`toast.success("Already authorized")` + revalidate the MCP row) rather than an error.

The Reauthorize button now passes `force=true` whenever the row is currently authorized; the initial Authorize button (no token yet) still posts without the flag.

## Test plan

- [x] Backend tests in `apps/backend/src/services/mcp-oauth-provider.test.ts` still pass.
- [x] Manual: with a valid refresh token, click Reauthorize → popup opens at the provider's authorize endpoint (force path).
- [x] Manual: without any tokens, click Authorize → identical behaviour to before (no `?force=true`).
- [x] Manual: trigger the silent-refresh path (page load reauthorize on a freshly-refreshed token) → success toast instead of failure toast.
- [x] No effect on Bearer / None auth types — the endpoint still rejects those upfront.

### DIFF
--- a/apps/backend/src/routes/mcp.test.ts
+++ b/apps/backend/src/routes/mcp.test.ts
@@ -1,12 +1,14 @@
 import { describe, it, expect, beforeEach, vi } from "vitest";
 import { mockDb, mockSession, resetMockDb } from "../test-utils.ts";
 import app from "../server.ts";
+import { auth as mcpAuth } from "@ai-sdk/mcp";
 
 vi.mock("@ai-sdk/mcp", () => ({
   experimental_createMCPClient: vi.fn().mockResolvedValue({
     tools: vi.fn().mockResolvedValue({ tool1: {} }),
     close: vi.fn().mockResolvedValue(undefined),
   }),
+  auth: vi.fn(),
 }));
 
 describe("MCP Routes", () => {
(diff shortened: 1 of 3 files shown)


willdady:
Nice fix — root cause writeup is spot on, and the `?force=true` opt-in plus `oauthClientId`/`oauthClientSecret` preservation is exactly right.

One ask before merging: could you add a route-level test for the `force=true` branch in `apps/backend/src/routes/mcp.ts`? The new conditional clears `oauthAccessToken` / `oauthRefreshToken` / `oauthTokenExpiresAt` / `oauthScope` and patches the in-memory `mcpRecord[0]` before invoking `mcpAuth` — none of that is covered by the existing `mcp-oauth-provider.test.ts` suite. A small test asserting:

1. `POST /mcps/:mcpId/oauth/authorize?force=true` nulls those four columns in the DB, and
2. the no-`force` path leaves them untouched

would lock in the behaviour and prevent a future refactor from quietly regressing the Reauthorize button back to its current state.

Happy to approve once that's in.

MesoX:
Thanks for the review! Added route-level tests in `apps/backend/src/routes/mcp.test.ts` covering both branches:

1. `POST /mcps/:mcpId/oauth/authorize?force=true` — asserts the DB update is issued and `oauthAccessToken` / `oauthRefreshToken` / `oauthTokenExpiresAt` / `oauthScope` are nulled, while `oauthClientId` / `oauthClientSecret` are explicitly *not* touched (locks in the client-credential preservation behaviour).
2. No-`force` path — `mcpAuth` returns `AUTHORIZED`, response is `{ alreadyAuthorized: true }`, and `mockDb.update` is never called.

Tests use the existing `test-utils` mock DB; `auth` from `@ai-sdk/mcp` is mocked alongside the existing `experimental_createMCPClient` mock, and the real `DatabaseOAuthClientProvider` is constructed by the route — the mocked `mcpAuth` invokes `provider.redirectToAuthorization(...)` to populate the pending URL, so we exercise the production code path end-to-end rather than stubbing the provider.

All 43 MCP tests (`mcp.test.ts` + `mcp-oauth-callback.test.ts` + `mcp-oauth-provider.test.ts`) pass. Pushed as a separate commit on the branch.

PR #146, opened 2026-05-28
feat(backend): let sandbox containers reach host services via host-gateway

MesoX:
## Summary

- The Docker reference sandbox adapter now injects `host.docker.internal:host-gateway` into `HostConfig.ExtraHosts` for every sandbox container at creation time
- On Linux, `host.docker.internal` is not automatically resolvable inside Docker containers (unlike Mac/Windows Docker Desktop) - without this, agent-written code cannot reach host-published services such as local model servers, transcription services, or any sidecar the operator runs on the host
- Operator-controlled via `PLATYPUS_SANDBOX_EXTRA_HOSTS` env var:
  - unset → default `host.docker.internal:host-gateway`
  - empty string → no ExtraHosts (opt out entirely)
  - `"name:ip,name2:ip2"` → exact list, replaces the default
- Entries are validated against Docker's accepted formats (hostname + IPv4/IPv6/`host-gateway`); a malformed entry fails sandbox provisioning loudly rather than silently
- Only affects newly-created containers; existing sandboxes keep their network config until destroyed and recreated (consistent with ADR-0003)
- ADR-0005 added documenting the threat-model delta: sandbox can reach operator-published host services; the `platypus-network` database and other internal services remain unreachable

## Test plan

- [x] Create a new sandbox on Linux - verify `host.docker.internal` resolves inside the container (`ping host.docker.internal` or `curl http://host.docker.internal:<port>`)
- [x] Set `PLATYPUS_SANDBOX_EXTRA_HOSTS=""` - verify no ExtraHosts are added
- [x] Set `PLATYPUS_SANDBOX_EXTRA_HOSTS="myhost:1.2.3.4"` - verify only that entry is present
- [x] Set a malformed value — verify sandbox creation fails with a clear error
- [x] Run existing Docker adapter tests: `pnpm --filter backend test`

### DIFF
--- a/apps/backend/.env.example
+++ b/apps/backend/.env.example
@@ -44,5 +44,18 @@ STORAGE_DISK_PATH=./data/files
 # Requires a reachable Docker daemon. See docs/adr/0003.
 PLATYPUS_SANDBOX_DOCKER_ENABLED=false
 
+# Sandbox host reachability — comma-separated ExtraHosts entries applied to
+# every Docker sandbox container at creation. Lets agent-written code reach
+# host-published services (e.g. local model servers, transcription, the
+# internal-resource-proxy) without hardcoded IPs. See docs/adr/0005.
+#
+#   unset           → host.docker.internal:host-gateway (default)
+#   empty string    → no ExtraHosts (opt out)
+#   "name:ip,..."   → exact list, replaces the default
+#
+# Only affects newly-created containers; existing sandbox containers keep
+# their config until destroyed and recreated.
+# PLATYPUS_SANDBOX_EXTRA_HOSTS=host.docker.internal:host-gateway
+
 # Frontend URL for generating resource links in tool responses
 FRONTEND_URL=http://localhost:3001
--- a/apps/backend/src/sandbox/backends/docker.test.ts
+++ b/apps/backend/src/sandbox/backends/docker.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeEach, vi } from "vitest";
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
 import { PassThrough } from "node:stream";
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


willdady:
Stepping back on the env var — I don't think a global knob is the right shape at all.

Concrete case: imagine an operator running two workspaces on the same host — one for a contractor, one personal. The personal workspace wants to reach a local Ollama and an internal proxy on the host; the contractor workspace should stay isolated. With a process-global env var there's no way to express that — both sandboxes get the same ExtraHosts or neither does. "What host services can this agent reach" is fundamentally a per-Sandbox capability, not a host-deployment setting.

Proposing we close this PR in favour of the redesign below.

## Shape

**Backend**
- Drop `PLATYPUS_SANDBOX_EXTRA_HOSTS` entirely.
- `dockerSandboxConfigSchema` gains:
  ```ts
  extraHosts: z.array(z.string().regex(EXTRA_HOST_PATTERN)).default([])
  ```
- Adapter reads `ctx.sandbox.config.extraHosts` at container create. New sandboxes start empty — no host reachability unless explicitly opted in.
- No migration concern: the docker adapter is already env-gated behind `PLATYPUS_SANDBOX_DOCKER_ENABLED`, and the schema was `{}` before.

**UI** (Sandbox edit page, docker adapter config panel)

```
Host reachability
  [ ] Allow host.docker.internal
      Lets agent-written code reach services on this host
      (local model servers, internal proxies, etc.). Off by default.

  > Advanced: custom host aliases
      [+ Add entry]  hostname        ip / host-gateway
```

The checkbox is shorthand for `["host.docker.internal:host-gateway"]`. Advanced disclosure exposes the full key/value list for the long tail. Mirrors the env-vars editor introduced in ADR-0004.

**Apply semantics**

ADR-0003 freezes container config at create time, so editing this on an existing sandbox can't take effect until recreate. Two options:

1. Show a "Pending — applies on next sandbox recreate" badge next to the field after save, with a "Recreate now" button.
2. Auto-destroy + recreate on save behind a confirm dialog.

I'd take (1) — consistent with how resource limits already behave and honest about ADR-0003's container-config-is-frozen rule. The explicit "Recreate now" button covers the operator who wants it immediate.

**ADR-0005**

Narrows to: "Host reachability is a per-sandbox capability declared on the docker adapter config. Default is none. Operators grant it explicitly per Sandbox via the UI / config column." The "DB remains unreachable" claim gets the caveat that it's a function of operator port publishing, not a structural guarantee.

Happy to pick this up if you'd rather not rework, or open it as a fresh issue and close this PR — your call.

MesoX:
Hi @willdady,

thanks for the review — while I agree on most things, I first iterated it as a UI configuration and found problems with this approach at the very same example you're presenting. I don't want to let the contractor choose that their sandbox accesses the whole network. That's why I went with the env way in this first version — it's much clearer that the admin is actually in charge of whether sandboxes (although all or none) can access Docker's internal network and services published there. If we put the workspace owner in charge of this, everyone can then access anything published by simply changing a switch button.

Walking through it:

**`extraHosts` can't express the contractor/personal split anyway.** `host.docker.internal:host-gateway` is all-or-nothing per sandbox: any service the operator publishes on the host is reachable, or none of them. There's no "WhisperX yes, MemPalace no" with this field — both are equally reachable the moment the entry is present.

**And the per-sandbox shape doesn't deliver isolation under current RBAC.** Workspaces are single-user; the workspace owner edits their own sandbox via `requireWorkspaceAccess`. So in the example, the contractor *is* the owner of their workspace and can flip `extraHosts` on themselves. The redesign moves the knob from the operator to the user it's supposed to restrict.

**What actually expresses the example: Docker networks + admin-only field.**

- **Deployment env var** declares the eligible set:
  ```
  PLATYPUS_SANDBOX_ALLOWED_NETWORKS=shared-services,public-tools,admin-only
  ```
  Operator runs WhisperX on `shared-services`, MemPalace on `admin-only`, etc. (compose-level).
- **`dockerSandboxConfigSchema`** gains `networks: string[]` (default `[]`). Adapter at container create uses the first entry as `HostConfig.NetworkMode` and post-create `network.connect()`s the rest. Submitted subset is validated against the allowlist server-side.
- **Field is org-admin-only.** Sandbox PUT rejects `networks` changes from non-admins with 403. Other fields (name, env) stay owner-editable. UI renders a multi-select fed by a new `GET /sandbox/networks` endpoint, disabled for non-admins.

Authority chain becomes honest: operator declares what exists, admin picks per-sandbox subset, contractor cannot escalate.

Threat coverage:
- Contractor edits own sandbox → field locked, 403. ✓
- Contractor crafts API call with `admin-only` → allowlist subset check rejects. ✓
- Admin misconfigures sandbox → blast radius bounded to operator-eligible networks. ✓
- Operator retires `admin-only` → remove from env, new sandboxes can't attach (existing containers keep config until recreated, per ADR-0003). ✓

Decisions on the smaller knobs:
- **Default for new sandboxes:** `[]`. Default-deny matches the principle that motivated the redesign.
- **Migration:** none. PR #146 not shipped yet, allowlist defaults to empty, vanilla installs get no networks attached.
- **Loss of `host.docker.internal` default:** acceptable. Operators relying on it run host services as containers on a shared network instead — that's the pattern the new shape encourages anyway.

One open question I'd want your call on before cutting code:

**Strict super-admin or org-admin?** `orgMembership.role === "admin"` allows multiple admins per org. Comfortable with any of them editing this field, or should it be platform-level super-admin only? Tighter is safer for multi-tenant deployments where org admin might be a customer-of-customer, looser is friendlier for self-hosted single-org installs.

If you're sold on the direction, happy to close #146 and open a fresh PR for `networks`. If you'd rather keep #146 narrow (env-only `extraHosts`) and tackle networks as separate work, also fine — they're independent.

willdady:
Closing in favour of #151, which implements the direction we landed on in this thread.

Thanks @MesoX — your RBAC analysis here was the crux of it: because workspaces are single-user, a per-Sandbox reach field guarded by `requireWorkspaceAccess` would hand the knob to the contractor it's meant to restrict. #151 takes your `networks` + operator-allowlist + admin-only design and generalises it: reach fields (`networks`, `extraHosts`) are org-admin-only and default-deny, with the operator declaring the eligible set via `PLATYPUS_SANDBOX_DOCKER_ALLOWED_NETWORKS`. It also folds in a broader governance model (ADR-0006) and a two-tier sandbox env split (ADR-0004 amendment). The decisions are written up in ADR-0005/0006.

PR #147, opened 2026-05-28
feat(sub-agent): configurable per-relationship output context budget

MesoX:
## Summary

Sub-agents that communicate via shared state (files, board cards, DBs) shouldn't dump their entire response text back into the parent agent's context window. With long-running tools (e.g. transcript producers, Drive listers, API proxies) this quickly exhausts the model's context with payloads the parent never reads.

This PR adds a per-relationship setting controlling how much of a sub-agent's response is returned to the parent's context:

- **`full`** — complete response (default, backwards-compatible)
- **`none`** — returns the sentinel `"Task completed."` only
- **`<integer>`** — returns the last N characters (where agents typically place summaries/status lines)

The setting lives on the parent → sub-agent **edge**, not on the sub-agent itself, so the same sub-agent can be reused from multiple parents with different context budgets.

## Why per-edge, not per-agent

A `librarian` sub-agent might want to return a full result to one parent (which acts on it) but only a summary line to another parent (which orchestrates many tasks). Storing the budget on the sub-agent record would conflate those cases. Also there might be specific cases (like when using Sandbox environment), where you can store results in a file if needed (e.g. in case the result is big). Most prevalent use case is system doing a full transcript of a long meeting - I don't need whole transcript being handed over from one sub-agent to main agent and then handed back to another sub-agent who should extract it. All this just pollute context without having any actual benefit.

### Schema

- `agent.sub_agent_ids` changes from `text[]` to `{ id, parentOutput? }[]` (JSONB column unchanged; only the value shape changes)
- Migration **0036** idempotently converts existing string-array rows to the new object-array shape — only rewrites rows whose first element is still a plain string, so it's safe to re-run

### Backend

- `createSubAgentTool` gains `parentOutput?: 'full' | 'none' | number` (default `'full'`); `toModelOutput` applies the configured truncation
- `parseParentOutput()` defensively handles `null` / `undefined` / non-numeric strings
- `chat-execution.ts` merges per-edge config into each sub-agent record before instantiating tools
- `validateSubAgentAssignment` updated for the new ref shape

### Frontend

- The Sub-Agents card in the agent form now shows a per-toggle dropdown when a sub-agent is enabled
- "Last N chars" reveals a number input where the limit can be set explicitly
- New `OutputReturnedRow` component encapsulates the select+input pair

### Tests

- `parseParentOutput` — null/empty/zero/negative/garbage edge cases
- `toModelOutput` — all three modes (`'full'` / `'none'` / number) plus fallbacks
- `validateSubAgentAssignment` tests updated to the new ref-object signature

## Migration safety

The data migration is **idempotent**: it only rewrites rows where the first element of `sub_agent_ids` is a JSON string. Running it twice (or on a fresh DB created from this schema) is a no-op.

## Possible questions

I am honestly not 100 % sure how to handle this UI/UX wise. I have tried to make it work with two column as you have originally had in there, but adding two fields under each sub-agent is pretty hard.

## Test plan

- [x] `pnpm test` passes
- [x] On a DB with legacy string-array `sub_agent_ids`, `pnpm drizzle-kit-migrate` converts the data without errors
- [x] In the UI, toggling a sub-agent on shows the dropdown; switching to "Last N chars" reveals the number input
- [x] After saving, the budget is honored: with `'none'`, the parent sees only `"Task completed."`; with `500`, it sees the trailing 500 chars
- [x] Existing agents without any configured output budget continue to return their full response (no behavior change)

### DIFF
--- a/apps/backend/drizzle/0036_relationship_subagent_output.sql
+++ b/apps/backend/drizzle/0036_relationship_subagent_output.sql
@@ -1,0 +1,29 @@
+-- Convert agent.sub_agent_ids from a flat string array of sub-agent ids:
+--    ["id1", "id2"]
+-- to an object array carrying per-relationship configuration:
+--    [{"id": "id1"}, {"id": "id2"}]
+--
+-- This migration is required because parentOutput (how much of a sub-agent's
+-- response text is returned to the parent's context window) is now configured
+-- per parent→sub-agent edge rather than as a global property of the sub-agent.
+-- Storing the config alongside the id lets the same sub-agent be reused from
+-- multiple parents with different context budgets.
+--
+-- Idempotent: only rewrites rows whose first element is still a plain string.
+
+UPDATE "agent"
+SET "sub_agent_ids" = (
+  SELECT COALESCE(
+    jsonb_agg(
+      CASE
+        WHEN jsonb_typeof(elem) = 'string' THEN jsonb_build_object('id', elem #>> '{}')
+        ELSE elem
+      END
+    ),
+    '[]'::jsonb
+  )
+  FROM jsonb_array_elements("sub_agent_ids") AS elem
+)
+WHERE "sub_agent_ids" IS NOT NULL
+  AND jsonb_array_length("sub_agent_ids") > 0
+  AND jsonb_typeof("sub_agent_ids" -> 0) = 'string';
--- a/apps/backend/drizzle/meta/0036_snapshot.json
+++ b/apps/backend/drizzle/meta/0036_snapshot.json
@@ -1,0 +1,3641 @@
+{
+  "id": "8cd05ba1-a597-46c3-9083-f858a676e8f9",
+  "prevId": "cf2b5eda-7cbc-4df9-8a06-c991ebf1f980",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.agent": {
+      "name": "agent",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "workspace_id": {
+          "name": "workspace_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "provider_id": {
+          "name": "provider_id",
+          "type": "text",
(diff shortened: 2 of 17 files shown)


willdady:
Thanks for the work here — the migration and tests are solid. But I don't think we should take this change, and I want to explain why, because it hinges on a detail about how sub-agents already work that I suspect isn't obvious.

**A sub-agent only ever returns its final message to the parent.** From `createSubAgentTool`:

```ts
// the only thing the parent receives:
yield { entries, text: await result.text } satisfies SubAgentActivity;

toModelOutput: ({ output }) => ({
  type: "text",
  value: (output as SubAgentActivity)?.text ?? "Task completed.",
});
```

`result.text` is the sub-agent's **final assistant message** — and nothing else. The full transcript, every intermediate tool result, the reasoning steps, the giant Drive listing or meeting transcript all live in `entries`, which is consumed solely by the UI activity stream and is **never handed to the parent model.** None of it crosses the boundary.

So the motivating scenario in the description — *"a full transcript of a long meeting being handed over from one sub-agent to the main agent"* — does not happen today. The transcript reaches the parent only if the sub-agent **chooses to echo it in its final message.** What this PR actually controls is not "the sub-agent's output," it's "how verbose the sub-agent's closing line is."

That is already controllable, with zero schema or UI changes, through the parent agent's system prompt — which the workspace user authors. They can write, *"When delegating to Agent X, always instruct it to reply with only 'Task complete' as its final message,"* and the parent will bake that into each `task` it sends. That gives the exact per-relationship behavior this PR is built around, dynamically, for free.

(To be clear: we can't solve this on the sub-agent side instead, because an agent and a sub-agent are the same entity — any agent can also be used as a top-level, human-interactive agent, so we can't assume it should always close with a terse status line. The instruction belongs on the calling side, which is precisely where the system prompt already lets us put it.)

The only thing prompting can't provide is a hard, model-independent guarantee, and `last N chars` is strictly worse than just asking the model to summarize. That doesn't justify a new data shape on a core table, a migration, and a UI surface for a problem the system prompt already solves.

I'm going to close this. The real gap, if any, is documentation — we should make it clear that sub-agents return only their final message and that delegation verbosity is controlled from the parent's prompt. Appreciate the effort, and the migration pattern here is a good reference for future work.

PR #168, opened 2026-05-31
feat(org): configurable agent-run timeouts with env-clamped ceiling

MesoX:
## Summary

Adds a per-organization override of agent-run wall-clock + per-step timeouts for both chat-driven and trigger-driven runs. Defaults are unchanged; overrides only take effect when an org admin sets them in the organization settings UI. Every override is clamped server-side to a deployer-supplied environment ceiling — an admin can lower a timeout but never raise it past what the host operator allows.

## Motivation

Long-running agent runs (multi-step research, slow MCP/tool calls) sometimes need more headroom than the hardcoded defaults, but the limit should stay under the deployer's control. This exposes a safe, bounded knob per organization.

## Changes

**Schema**

- Migration `0041` adds a nullable `jsonb agent_run_settings` column on `organization` (idempotent `ADD COLUMN IF NOT EXISTS`; existing rows keep `null` and behave exactly as before).
- `AgentRunSettings` exported from `@platypus/schemas` as a strict zod object with four optional positive-int fields.

**Backend**

- New `services/agent-run-settings.ts` — `resolveRunTimeouts(orgId, kind)` reads the org override, falls back to env, then to hardcoded defaults, and clamps to the env ceiling. Chat defaults are sourced from `run-registry` constants so they cannot drift.
- `PUT /organizations/:orgId` (admin-only) rejects overrides above the env ceiling with `400 { error }` naming the offending env vars.
- New `GET /organizations/:orgId/agent-run-settings/ceilings` exposes the current ceilings for the UI.
- `routes/chat.ts` and `services/trigger-execution.ts` resolve timeouts via the new service instead of reading env directly.

**Frontend**

- `OrganizationForm` gains an "Agent run timeouts" section (edit only) with four minute-valued inputs, ceiling placeholders, client-side ceiling validation for inline feedback, and surfacing of the server error message.

**Config**

- `.env.example` documents the four ceiling env vars and their defaults (`RUN_PER_RUN_TIMEOUT_MS`, `RUN_PER_STEP_TIMEOUT_MS`, `TRIGGER_PER_RUN_TIMEOUT_MS`, `TRIGGER_PER_STEP_TIMEOUT_MS`).

## Behavior compatibility

Chat defaults (10 min run / 2 min step) are imported from `run-registry`'s existing defaults, so a run with no env override and no org override behaves identically to before this PR.

## Testing

- `services/agent-run-settings.test.ts` — env defaults, env overrides, garbage-env rejection, DB-backed resolution including ceiling clamping, no-row fallback, chat-vs-trigger isolation.
- `routes/organization.test.ts` — PUT path: 403 for non-admins, 400 (with `error` key) when over the ceiling, successful within-ceiling persist, and null clears the override.
- Full suite: backend 795 pass, frontend 16 pass.


MesoX:
Hey @willdady this would deliberately help with #261 - Chat compaction can take some time (even more on the locally used models on slower hardware (not a big problem to see even 5 minutes of compaction happening with long context and slow dense model). Even with faster model we have run into timeouts. I would update it to the latest main, just let me know whether this is something you want to add to the tool tiself, or is there any other way how to make these things a bit more configurable?

willdady:
Closing this, @MesoX — with an apology attached, because you asked me a direct
question in July ("is this something you want in the tool itself, or is there
another way to make these configurable?") and I never answered it. That one's on
me, and it's the reason this sat long enough to go stale.

I'm not going to answer it with a yes or no in this thread, though — that's the
part I got wrong by leaving it open. Whether Platypus should expose run timeouts
as configuration, and at what scope, is a design question that deserves proper
triage rather than a decision made under the pressure of an open PR that already
implements one particular answer.

On the branch itself: main has moved underneath it regardless. `0047` is now
`0047_backfill_model_configs.sql` (we're at `0051`), and #487 rerouted Sub-Agent
runs through the run lifecycle so a delegated run inherits `parentRun.timeouts`
rather than resolving its own — which changes a question this PR never had to
answer.

**Please raise it as an issue.** Lead with the problem rather than the
implementation: the runs that hit the ceiling, the model and hardware, the
observed durations, what failed. The 5-minute compaction case on a slow local
model is good material, and it's stronger stated generally — a slow MCP tool or
a long tool result hits the same wall with no compaction involved (compaction
itself is back at the issue stage, see #261).

Things I'd want the issue to work through before any code, none of them decided:

- **Scope.** Org-level is one option. A slow model is arguably a property of the
  Provider or the Agent rather than the Organization, and it's not obvious which
  should own the knob.
- **Whether it's user-facing configuration at all.** Trigger runs are already
  operator-configurable via `TRIGGER_PER_STEP_TIMEOUT_MS` /
  `TRIGGER_PER_RUN_TIMEOUT_MS`, while Chat runs are hardcoded at
  `run-registry.ts:65`. Whether that asymmetry is a gap to close or a boundary
  to keep is worth arguing explicitly — on a self-hosted deployment the person
  who needs a longer timeout usually has shell access to the box.
- **Sub-Agent inheritance.** Post-#487, does a delegated run inherit the
  parent's resolved timeout or re-resolve its own?
- **Ceilings.** The env-clamped ceiling you built is a sound instinct, but
  whether a separate ceiling is needed depends on where the setting lands.

The work here isn't wasted — it's a concrete proposal to point the issue at, and
the branch stays readable after it's closed. I'd just rather settle the shape
before another migration goes in.

Thanks for the patience, and sorry again for the silence on this one.

PR #266, opened 2026-06-20
fix(backend): default chat-mode baseUrl to OpenAI's endpoint

willdady:
## Summary

Follow-up to #262. That PR routed `apiMode=chat` OpenAI providers through `@ai-sdk/openai-compatible`, but `baseUrl` is optional in the provider schema. A real-OpenAI provider set to chat mode (no `baseUrl`) was building the compatible client with an empty `baseURL`, so every request targeted an empty URL and failed — the OpenAI SDK had previously defaulted this to `api.openai.com`.

## Changes

- `provider.ts`: fall back to `https://api.openai.com/v1` when no `baseUrl` is configured. Uses `||` (not `??`) so the empty-string case the form can produce is also caught.
- Adds a regression test for chat mode without a `baseUrl`.

## Testing

- `provider` unit tests (62 passed), `typecheck`, and `lint` pass.

### DIFF
--- a/apps/backend/src/services/provider.test.ts
+++ b/apps/backend/src/services/provider.test.ts
@@ -122,6 +122,20 @@ describe("openProvider", () => {
     expect(mockCreateOpenAI.creator).toHaveBeenCalled();
   });
 
+  it("falls back to OpenAI's endpoint for chat mode without a baseUrl", () => {
+    // baseUrl is optional, so a real-OpenAI provider in chat mode arrives with
+    // it empty. The compatible client must still target OpenAI rather than an
+    // empty URL.
+    openProvider({
+      ...baseProvider,
+      apiMode: "chat",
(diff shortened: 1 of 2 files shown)


MesoX:
@willdady I am not sure here if we are not trying to reinvet the wheel - most similar tools have OpenAI and OpenAI Compatible as two separate provider type options simply because there are numerous things which are different (e.g. mandatory base url, API typa, SDK and so on). It would possibly be for the good of the Platypus to split it up in the future too.

willdady:
@MesoX base URL still needs to be configurable for **both** the 'official' OpenAI provider and the OpenAI-compatible provider. Need to consider the user may need to reach the official API via some other proxy (different base URL).

That said, I'm not opposed to making the split more explicit if there's a more concrete reason to do so. Feel free to open a separate issue if you wish.

PR #349, opened 2026-07-18
fix: centralize file classification

### LINKED ISSUE #347: Deduplicate the three-way file classifier across backend and frontend
## Summary

The passthrough / text / reject file-classification cascade (issue #328) is implemented **twice** — once on each side of the wire — and the two copies can silently drift, even though the design explicitly promises they won't.

Surfaced during the pre-release audit of 2.1.4 (#343/#344/#345). **Non-blocking** — the two copies agree today; this is a post-release cleanup to remove the drift risk.

## The duplication

**Backend** — `apps/backend/src/services/file-classification.ts` (`classifyFilePart`):
```ts
if (mediaTypeMatches(part.mediaType, passthroughFileTypes)) return "passthrough";
if (isTextLikeExtension(part.filename) && !(bytes && looksBinary(bytes))) return "text";
return "reject";
```

**Frontend** — `apps/frontend/lib/model-config.ts` (`classifyAttachment`):
```ts
if (mediaTypeMatches(file.mediaType, passthroughFileTypes)) return "passthrough";
if (isTextLikeExtension(file.filename)) return "text";
return "reject";
```

Same cascade, same order, same return union. The only intended difference: the backend takes an optional `bytes` arg and adds a `looksBinary` content-sniff to the `text` branch; the frontend has no bytes at classification time, so it omits that guard.

## Why it matters

`file-classification.ts` states the whole point of the shared module:

> "The gate and the send-time normalizer share this logic so the reject decision and the transform decision never drift."

That guarantee holds **within** the backend (both the gate and the normalizer call `classifyFilePart`), but the **frontend warning** — `classifyAttachment`, which drives the `file-compatibility-warning.tsx` copy — is a hand-copied third implementation of the same cascade.

If someone later changes the classification rule (reorder passthrough-vs-text, add a fourth outcome, make text-detection also consult the media type), they'd update the backend and the frontend warning would quietly disagree — e.g. the UI says "will be inlined as text" while the backend actually rejects. That's a confusing UX bug, not a crash, which is exactly the kind of drift that slips through review.

**Mitigating factor (why it's low priority):** the primitives are already shared — both copies import `mediaTypeMatches` and `isTextLikeExtension` from `@platypus/schemas`. Only the ~6-line combining cascade is duplicated, so the drift surface is small and the two copies are currently consistent.

## Proposed fix

Lift the metadata-only cascade into `@platypus/schemas` (where the primitives already live) as the single source of truth, and have both sides call it:

```ts
// packages/schemas/index.ts
export const classifyFile = (
  file: { mediaType?: string; filename?: string },
  passthroughFileTypes: string[],
  looksBinary?: boolean,   // backend passes a computed flag; frontend omits
): "passthrough" | "text" | "reject" => {
  if (mediaTypeMatches(file.mediaType, passthroughFileTypes)) return "passthrough";
  if (isTextLikeExtension(file.filename) && !looksBinary) return "text";
  return "reject";
};
```

- Backend `classifyFilePart` becomes a thin wrapper that computes `looksBinary(bytes)` and delegates (keeping `bytes`/sniff on the backend where bytes are available).
- Frontend `classifyAttachment` calls `classifyFile` with the flag omitted.

The `bytes`/sniff asymmetry is preserved by the optional `looksBinary` param; the drift risk goes away.

## Acceptance criteria

- [ ] A single classifier function is the source of truth (in `@platypus/schemas`).
- [ ] Backend gate + normalizer and the frontend warning all delegate to it.
- [ ] Backend still applies the `looksBinary` content-sniff; frontend behavior unchanged.
- [ ] `pnpm typecheck` / `pnpm lint` / `pnpm test` pass.

floze-the-genius:
## Summary

Closes #347.

Moves the shared passthrough / text / reject cascade into
`@platypus/schemas` so the frontend warning and backend enforcement use one
classifier.

## Changes

- adds `classifyFile` and `FileClassification` to the shared schemas package
- keeps the backend-only NUL-byte sniff in `classifyFilePart`
- delegates frontend metadata classification directly to the shared function
- adds shared regression coverage for passthrough priority, text files, binary
  sniffing, and unsupported files

The current behavior is preserved: backend bytes can override a text-like
extension, while native passthrough still wins before content sniffing.

## Testing

- `pnpm --filter @platypus/schemas test` - 48 tests passed
- `pnpm --filter @platypus/backend test -- file-classification.test.ts` -
  1,174 backend tests passed
- `pnpm --filter @platypus/frontend test` - 23 tests passed
- `pnpm typecheck`
- `pnpm --filter @platypus/frontend lint`
- Prettier check for all changed TypeScript files
- `git diff --check`

### DIFF
--- a/apps/backend/src/services/file-classification.ts
+++ b/apps/backend/src/services/file-classification.ts
@@ -15,14 +15,19 @@
  * browser-supplied media type, which is an unreliable, OS-specific lottery
  * (`.py` → `text/x-python` but `.md` → `application/octet-stream`).
  */
-import { isTextLikeExtension, mediaTypeMatches } from "@platypus/schemas";
+import {
+  classifyFile,
+  isTextLikeExtension,
+  mediaTypeMatches,
+  type FileClassification,
+} from "@platypus/schemas";
 
 // The media-type matcher and the text-extension test live in @platypus/schemas
 // so the backend gate and the frontend warning share one copy. Re-exported so
 // this module remains the single classification import surface (incl. tests).
 export { isTextLikeExtension, mediaTypeMatches };
 
-export type FilePartClass = "passthrough" | "text" | "reject";
+export type FilePartClass = FileClassification;
 
 /** How many leading bytes to sniff for the NUL-byte binary heuristic. */
 const SNIFF_BYTES = 8_000;
@@ -46,11 +51,9 @@ export const classifyFilePart = (
   passthroughFileTypes: string[],
   bytes?: Uint8Array,
 ): FilePartClass => {
-  if (mediaTypeMatches(part.mediaType, passthroughFileTypes)) {
-    return "passthrough";
-  }
-  if (isTextLikeExtension(part.filename) && !(bytes && looksBinary(bytes))) {
-    return "text";
-  }
-  return "reject";
+  return classifyFile(
+    part,
+    passthroughFileTypes,
+    bytes ? looksBinary(bytes) : false,
+  );
 };
--- a/apps/frontend/lib/model-config.ts
+++ b/apps/frontend/lib/model-config.ts
@@ -1,7 +1,6 @@
 import {
+  classifyFile,
   defaultPassthroughFileTypes,
-  isTextLikeExtension,
-  mediaTypeMatches,
   type Provider,
 } from "@platypus/schemas";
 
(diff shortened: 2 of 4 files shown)


willdady (changes requested):
One small tidy before merge:

**`apps/frontend/lib/model-config.ts` (module docstring, ~lines 8–14)** still lists `mediaTypeMatches` and `isTextLikeExtension` as "the pure logic … shared with the backend," but this PR removed those imports in favor of `classifyFile`. Please update the comment to name `classifyFile` so it tracks the actual imports.

Everything else looks good — clean dedup, behavior preserved, well-tested. Happy to approve once this is fixed.

floze-the-genius:
Updated the module comment to reference the actual shared helpers (`defaultPassthroughFileTypes` and `classifyFile`) in commit 345f8dc. Prettier and `git diff --check` pass.

willdady (approved):
Docstring tidy looks good — `model-config.ts` now names `classifyFile` and the stale `mediaTypeMatches`/`isTextLikeExtension` references are gone.

Clean dedup, behavior preserved across both call sites, well-tested. LGTM 🚀

Issue #329, opened 2026-07-14
Feature: pluggable web-search/browse backend routed through the search toggle


MesoX:
# Feature: pluggable web-search/browse backend routed through the search toggle

## Summary

Add a new plugin **extension point** — a **web backend** — that the existing "web search"
chat toggle routes to. A web backend contributes a small, curated pair of tools (a `web_search`
and a `read_url`), injected **ephemerally** only when the toggle is on, exactly like today's
native provider search. This lets self-hosted / OpenAI-compatible deployments (local vLLM,
LiteLLM proxies) get first-class web search + page reading — backed by whatever the operator runs
(e.g. SearXNG for search + [obscura](https://github.com/h4ckf0r0day/obscura) for browsing) —
without baking any vendor specifics into core, and without the always-on tool-schema bloat that
connecting those services as MCP servers causes.

Selection mirrors the existing **sandbox backend** model (a `backend` id chosen per-provider /
per-workspace). This is deliberately a new extension point, not a normal tool set: it feeds the
**request-gated search slot**, not the always-on agent toolbox.

## Prior discussion / why revisit now

Self-hosted web search came up before, and the recommendation at the time was to do it via **MCP**
— which was the right call then: the plugin system did not exist yet, so MCP was the only
extension mechanism available, and "talk to an external service over a protocol" is squarely MCP's
job. That guidance still holds for the general case.

What changed is the introduction of the plugin system (ADR-0013) and, specifically, two things it
brought that MCP does not offer here:

1. a **curated in-process contribution surface** where core can own a slot and a plugin fills it,
   and
2. the precedent of **swappable, deployment-chosen backends** (sandbox backends).

Those are exactly what a toggle-driven, minimal, swappable web capability needs — and neither
existed at the time of the earlier MCP recommendation. So this isn't reopening a settled call; it's
noting that the extension mechanisms available have grown, and a path that wasn't possible then is
now the cleaner fit. MCP remains perfectly valid for anyone who just wants to wire a search service
in today (see Alternatives).

## Motivation

Web search today is **provider-native only**. `getOpenProvider` wires a `searchTools()` for
OpenAI / Anthropic / Google / OpenRouter; the OpenAI-**compatible** branch (local vLLM, LiteLLM)
returns none. So for a self-hosted model the "web search" toggle injects nothing — and because
`canSearch` today only excludes Bedrock, the button is still offered by default (the operator has
to disable it per-provider via `nativeSearchEnabled` to hide the dead switch).

[...]

PR #384, opened 2026-07-27
feat: add the web-search backend Extension point

MesoX:
Implements the plumbing layer of [ADR-0014](docs/adr/0014-web-search-backend-extension-point.md) — the
fourth core-owned Extension point, filling the request-gated web-search toggle for Providers without
working native search (self-hosted vLLM / LiteLLM / SGLang). Part 1 of 4 for #329; **no chat behaviour
changes yet** — nothing resolves a backend until PR2 adds the `provider.webBackend` column and the
turn-time gate.

Builds on #358 (the shared egress guard) and #357 (the amended ADR).

## What's here

- **`packages/plugin-sdk`** — `webBackends?` on `PluginContributions`, plus `WebBackendContribution`,
  `WebBackendExecutors`, `WebBackendContext`, `WebSearchResults`, `ReadUrlResult`. Append-only, so
  `PLUGIN_API_VERSION` stays **1** and `OLDEST_SUPPORTED_API_VERSION` is untouched; package `0.1.1 →
0.2.0`, README gains a "Web-search backends" section. (Publishing stays your release step.)
- **`apps/backend/src/web-backends/`** — the registry (a direct mirror of `sandbox/index.ts`), the
  core-owned limits as named exported constants, and `composeWebBackend`: the wrapper that turns a
  contribution into `{ backend, name, buildTurnTools(ctx) }`.
- **Loader** — registration loop with the usual id namespacing (core flat, third-party `${name}.${id}`)
  and plugin-attributed collision errors; `registerWeb?` injection point; `webBackendIds` on
  `LoadedPlugin`.
- **Registry / `GET /plugins` / boot log** — each plugin's web backends are reported like its tool sets
  and sandbox backends.

A backend supplies **executors**, not `Tool`s. Core owns the input schemas, the descriptions, the caps,
the slicing, the timeout, the error contract, and the egress guard — per the ADR's enforceability
argument: a backend that owned the `Tool` would put the model-supplied URL out of core's reach.

`read_url` mirrors `fetchUrl` exactly — same `max_length` / `start_index` defaults, same
`{ content, url, content_type, truncated, next_start_index? }` return, same trailing
`[Content truncated. Pass start_index=… to continue reading.]` hint — so a continuation read reads
identically whichever page-reader the model reached for. No robots.txt, per the ADR.

One case has no `fetchUrl` counterpart to mirror: `fetchUrl` has no core content cap, so nothing there
corresponds to reaching the tail of a page that `MAX_READ_URL_CONTENT_CHARS` already cut. Left alone,
that read returns `truncated: true` with no `next_start_index` — an unexplained dead end. Core emits a
distinct terminal line instead, saying the remainder cannot be read. The paging hint above is unchanged
and still byte-identical.

## Calls the ADR left open

Each is one constant or one helper, so revising any of them later is a one-line edit plus a test.

| Decision                                                                                                                                                                  | Why                                                                                                                                                                                                                                                                                                                   |
| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `MAX_SEARCH_RESULTS=10`, `MAX_SNIPPET_CHARS=500`, `MAX_TITLE_CHARS=200`, **`MAX_ANSWER_CHARS=4000`**                                                                      | The ADR names result-count capping and snippet truncation, which left `answer` — free text from Brave/Tavily — unbounded, reopening the context-blowup hole `createExecutors` exists to close.                                                                                                                        |
| **`MAX_SEARCH_RESULT_SCAN = MAX_SEARCH_RESULTS * 10`** — how many raw entries core will look at to fill its 10 slots, with a warn-log naming what was ignored             | The result cap alone does not bound the work: unusable entries are dropped and skipped _past_, so an executor returning a million `javascript:` hits would have core parse a million URLs. Same category as the two bounds above — a limit the ADR is silent on, fixed as a named constant rather than left implicit. |
| **`MAX_READ_URL_CONTENT_CHARS=1_000_000`**, applied before slicing                                                                                                        | Mirrors sandbox `MAX_READ_BYTES`. Stated limit: the executor hands core a materialised string, so this bounds context and CPU, not backend memory — the timeout is the only lever there.                                                                                                                              |
| **`timeoutMs` ceiling 120_000, refused at boot** (default 30_000)                                                                                                         | ADR wording is "refuses anything above a hard ceiling"; `timeoutMs` is static on the contribution, so it is knowable at load, and boot posture is fail-loud everywhere else. Not a silent clamp.                                                                                                                      |
| **Result URLs are `http(s)`-only** (`isPresentableUrl`), non-conforming entries dropped with a warn-log                                                                   | The egress guard covers URLs going _into_ `read_url`; nothing covered the ones coming _out of_ `web_search`. PR3 makes these clickable, and `javascript:`/`data:` in a pill is a real hole.                                                                                                                           |
| **Mandatory `web_search` enforced twice** — boot rejects a missing `createExecutors`; turn time warns and serves `{}` if the executor object has no `web_search`          | The TS type makes it required, but a third-party JS plugin can return `{}`. Boot fail-loud, runtime graceful — the ADR's split posture.                                                                                                                                                                               |
| **`buildTurnTools(ctx)`**, not `createTools(ctx)`                                                                                                                         | `createTools` is the identifier the ADR **rejects** for the contribution factory; reusing it on core's side would read as the rejected shape.                                                                                                                                                                         |
| **Casing seam:** SDK types camelCase (`contentType`), model-facing returns snake_case (`content_type`)                                                                    | SDK types follow repo convention; the return must match `fetchUrl` byte-for-byte, and the tool _names_ are already a documented snake_case exception.                                                                                                                                                                 |
| **Observability** is a defined line: `{ backend, plugin, tool, durationMs, outcome }`, `outcome ∈ ok \| timeout \| blocked \| error` — `debug` for `ok`, `warn` otherwise | Queries and URLs are user content and can carry sensitive terms, so they are logged at `debug` only, never on the outcome line.                                                                                                                                                                                       |
| **Model-facing failure text is fixed, not the upstream cause**                                                                                                            | A backend's error message routinely embeds the URL it called, which for a hosted search API carries the Operator's API key. The cause goes to the log line.                                                                                                                                                           |

Two seams worth a reviewer's eye:

- **`query` is echoed from the model's input**, not from the executor's returned `query` — core owns the
  result shape, so a backend cannot substitute text there. The field stays in the SDK type because the
  ADR defines the executor return that way.
- **`isPresentableUrl` lives in `web-backends/`** for now. PR3 needs the same check in the frontend
  before rendering a Sources pill; happy to lift it into `@platypus/schemas` there (or here, if you'd
  rather it land with the definition).

## Verification

- New: 34 tests in `web-backends/index.test.ts`, 11 in `plugins/loader.test.ts`, 2 in the SDK's
  `index.test.ts`. Covers the caps and the scan bound, the `javascript:` drop, egress blocking (real
  guard, injected resolver), slicing + both truncation messages, the content cap, timeouts, executor
  throws, boot refusals, the `fetchUrl`-mirroring schema defaults, and plugin-config injection.
- `pnpm typecheck`, `pnpm lint`, `pnpm test` all green across the workspace — backend 1320, frontend 26,
  schemas 60, plugin-sdk 7, example-plugin 3.

## Review pass on this branch

A self-review found nine items, all fixed here and all covered by the tests above. Three are worth a
reviewer knowing about, since they change behaviour rather than tidy it:

- **The contribution is passed to `composeWebBackend` by reference, never spread into a copy.** The
  namespaced id rides alongside it as a separate `backend` field. Spreading would invoke
  `createExecutors` on the copy, so a class-instance contribution would lose both its prototype and its
  `this` — the sandbox loop avoids the same trap by re-wrapping `create` around the original. Both
  executors are also bound to the object that supplied them, so a backend may write them as methods.
- **Executor payloads are read as `unknown` and narrowed field by field.** The SDK types say what a
  well-behaved backend returns; a third-party JS plugin is under no obligation to honour them. One
  visible consequence: a non-string `title` now narrows to `""` rather than being `String()`-coerced, so
  `[object Object]` cannot reach the model.
- **`backend` and `name` are validated before the id is namespaced.** Namespacing first would mint a
  plausible-looking `"acme.undefined"` and register it, turning a missing field into a mystery catalog
  entry instead of a boot error naming the plugin.

## Reporting surfaces — what this PR does and does not cover

`GET /plugins` reports `webBackends`, and the admin installed-plugins list renders them: its
`InstalledPlugin` type and its "No contributions." emptiness check both account for web backends, so a
plugin contributing only one no longer renders as contributing nothing.

The **docs site is deliberately not covered here.** `apps/docs/content/extending/index.mdx` documents
Sandbox backends and Tool sets and has no web-backend section, and `packages/plugin-sdk/README.md` links
to that page as the full contribution reference. That page describes what an Operator can actually use,
and until the Provider selector lands (PR3) there is no way to select a backend — so it goes with the
docs work in PR4 rather than describing an extension point nobody can reach. The SDK README, which ships
with the package an author builds against, _is_ updated here.

## Follows

PR2 predicate + `webBackend` column + turn-time resolution · PR3 catalog route, Provider selector,
`canSearch` rework, Sources rendering · PR4 ADR → `accepted` + docs, including the `extending` page's
web-backend section.

Refs #329 — part 1 of 4, so this does not close it.

## Related

#388 fixes the same two flaws in the Sandbox extension point — the `Object.prototype` registry and the
unvalidated contribution entry — which the review on this PR flagged as shared and asked to keep
separate. Based on `main`, independent of this branch.

### DIFF
--- a/apps/backend/index.ts
+++ b/apps/backend/index.ts
@@ -67,8 +67,9 @@ const main = async () => {
         origin: p.origin,
         toolSets: p.toolSetIds,
         sandboxBackends: p.sandboxBackendIds,
+        webBackends: p.webBackendIds,
       },
-      `Loaded plugin ${p.name}@${p.version} (${p.origin}): ${p.toolSetIds.length} tool set(s), ${p.sandboxBackendIds.length} sandbox backend(s)`,
+      `Loaded plugin ${p.name}@${p.version} (${p.origin}): ${p.toolSetIds.length} tool set(s), ${p.sandboxBackendIds.length} sandbox backend(s), ${p.webBackendIds.length} web backend(s)`,
     );
   }
   setLoadedPlugins(loadedPlugins);
--- a/apps/backend/src/plugins/loader.test.ts
+++ b/apps/backend/src/plugins/loader.test.ts
@@ -737,6 +737,51 @@ describe("loadPlugins — web-search backends", () => {
     ).rejects.toThrow(/@platypus\/searx.*"searx".*createExecutors/s);
   });
 
+  it("aborts (fail-loud) on a contribution with no backend id, before namespacing it", async () => {
+    const { register } = makeRegister();
+    // Namespacing an absent id first would mint `"@platypus/searx.undefined"`
+    // and register it, turning a missing field into a mystery catalog entry.
+    const nameless = {
+      name: "SearXNG",
+      createExecutors: () => ({
+        web_search: () => ({ query: "", results: [] }),
+      }),
+    } as unknown as ReturnType<typeof webBackend>;
+
+    await expect(
+      loadPlugins({
+        pluginNames: ["@platypus/searx"],
+        builtinPlugins: {
+          "@platypus/searx": () =>
+            Promise.resolve({
+              plugin: webManifest("@platypus/searx", [nameless]),
+            }),
+        },
+        register,
+        registerWeb: () => {},
+      }),
+    ).rejects.toThrow(/@platypus\/searx.*non-empty "backend" id/s);
+  });
+
+  it("aborts (fail-loud) on a contribution with a blank display name", async () => {
+    const { register } = makeRegister();
+    await expect(
+      loadPlugins({
+        pluginNames: ["@platypus/searx"],
+        builtinPlugins: {
+          "@platypus/searx": () =>
+            Promise.resolve({
+              plugin: webManifest("@platypus/searx", [
+                { ...webBackend("searx"), name: "   " },
+              ]),
+            }),
+        },
+        register,
+        registerWeb: () => {},
+      }),
+    ).rejects.toThrow(/@platypus\/searx.*"searx".*non-empty "name"/s);
+  });
+
   it("refuses an over-ceiling timeoutMs at boot rather than clamping it", async () => {
     const { register } = makeRegister();
     await expect(
--- a/apps/backend/src/plugins/loader.test.ts
+++ b/apps/backend/src/plugins/loader.test.ts
@@ -713,6 +713,33 @@ describe("loadPlugins — web-search backends", () => {
     ).rejects.toThrow(/@bad\/plugin.*webBackends.*array/s);
   });
 
+  it("aborts (fail-loud, plugin-named) on a non-object web backend entry", async () => {
+    // `webBackends` being an array is checked; a third-party JS plugin can
+    // still put `null` inside it, which would otherwise throw an unattributed
+    // "Cannot read properties of null" reading `.backend` next.
+    const { register } = makeRegister();
+    await expect(
+      loadPlugins({
+        pluginNames: ["@platypus/searx"],
+        builtinPlugins: {
+          "@platypus/searx": () =>
+            Promise.resolve({
+              plugin: {
+                name: "@platypus/searx",
+                version: "0.1.0",
+                apiVersion: 1,
+                contributes: { webBackends: [null] },
+              },
+            }),
+        },
+        register,
+        registerWeb: () => {},
+      }),
+    ).rejects.toThrow(
+      /@platypus\/searx.*web backend contribution must be an object/s,
+    );
+  });
+
   it("aborts (fail-loud) when a contribution has no createExecutors function", async () => {
     const { register } = makeRegister();
     // The TS type requires it; a third-party JS plugin can still omit it, and
--- a/apps/backend/src/plugins/loader.test.ts
+++ b/apps/backend/src/plugins/loader.test.ts
@@ -8,7 +8,12 @@ import {
   type SandboxBackend,
   type SandboxBackendContribution,
   type ToolSetContribution,
+  type WebBackendContribution,
 } from "@platypuschat/plugin-sdk";
+import {
+  MAX_WEB_TIMEOUT_MS,
+  type WebBackendRegistration,
+} from "../web-backends/index.ts";
 import { loadPlugins, parsePluginConfig, parsePluginList } from "./loader.ts";
 import { plugin as examplePlugin } from "./example/index.ts";
 import { registerToolSet, getToolSet } from "../tools/index.ts";
@@ -57,6 +62,34 @@ const toolSet = (id: string): ToolSetContribution => ({
   name: id,
   category: "Test",
   tools: {},
+});
+
+// A capturing `registerWeb` for the Web-search-backend extension point. It takes
+// core's composed registration, not the raw contribution.
+const makeWebRegister = () => {
+  const calls: WebBackendRegistration[] = [];
+  const registerWeb = (registration: WebBackendRegistration) => {
+    calls.push(registration);
+  };
+  return { registerWeb, calls };
+};
+
+const webManifest = (
+  name: string,
+  webBackends: WebBackendContribution[],
+): PlatypusPlugin => ({
+  name,
+  version: "0.1.0",
+  apiVersion: 1,
+  contributes: { webBackends },
+});
+
+const webBackend = (backend: string): WebBackendContribution => ({
+  backend,
+  name: backend,
+  createExecutors: () => ({
+    web_search: () => ({ query: "", results: [] }),
+  }),
 });
 
 const sandboxBackend = (backend: string): SandboxBackendContribution => ({
@@ -194,6 +227,7 @@ describe("loadPlugins", () => {
         origin: "core",
         toolSetIds: ["math-conversions", "time"],
         sandboxBackendIds: [],
+        webBackendIds: [],
       },
     ]);
   });
@@ -411,6 +445,7 @@ describe("loadPlugins — sandbox backends", () => {
         origin: "core",
         toolSetIds: [],
         sandboxBackendIds: ["docker"],
+        webBackendIds: [],
       },
     ]);
   });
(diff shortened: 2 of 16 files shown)


willdady (changes requested):
Read this against ADR-0014, `tools/fetch.ts`, `sandbox/index.ts`, and the egress guard. The core-owns-the-`Tool` split is implemented faithfully, the ADR's open calls are all pinned to named constants, and the seams you flagged for a reviewer are the right ones — the no-spread/`bind` handling, validate-before-namespace, and echoing `query` from the model's input all hold up. `read_url` does mirror `fetchUrl` byte-for-byte, including the extra terminal line for the capped tail.

Six things below. One I'd call blocking; the rest are small and each closes something this PR already set out to close.

## Blocking

**`createExecutors` is unguarded at turn time** — `web-backends/index.ts:415`

```ts
buildTurnTools: async (ctx) => {
  const executors = await contribution.createExecutors(ctx, plugin);
```

No `try/catch`, no `withTimeout`. Two consequences that contradict claims made elsewhere in the PR:

- A factory that throws (lazy credential validation, constructing a client against a dead endpoint) rejects `buildTurnTools`, so PR2's caller inherits an exception — where the missing-`web_search` case two lines down deliberately degrades to `{}`. ADR-0014's "runtime resolution graceful" consequence covers the factory too, not just the executors.
- A factory that _hangs_ pins the turn open. `timeoutMs` guards executor calls only, so "so a backend cannot pin a turn open" (loader comment, and the README) isn't true of the factory.

`withTimeout(() => contribution.createExecutors(ctx, plugin), timeoutMs)` inside a `try/catch` that warns and returns `{}`, plus two tests.

## Worth fixing before merge

**The scan-bound warn fires when the _result cap_ ended the loop** — `web-backends/index.ts:301`

`if (entries.length > scanned.length)` doesn't distinguish "the scan bound cut the tail" from "10 good results filled the slots and we broke at index 10". A backend returning 500 usable results (Brave and Tavily both default well above 10) gets this warn on every single search, telling the Operator a tail was ignored by a bound the loop never reached. Gate it on the loop exhausting `scanned` while still under `MAX_SEARCH_RESULTS`.

**`results[].url` and `content_type` escape the caps** — `web-backends/index.ts:288`, `:399`

Title, snippet, `answer`, and `content` are all bounded; these two are passed through at whatever length arrived, and `isPresentableUrl` only checks the scheme. So 10 × multi-megabyte `https://…` URLs reach the context window through the same path `MAX_ANSWER_CHARS` exists to close, and PR3 renders them as pills. Suggest a `MAX_URL_CHARS` that _drops_ the entry rather than truncating (a cut URL is a broken link, worse than no link), and a cap on `content_type`.

**The registry inherits `Object.prototype`** — `web-backends/index.ts:57-62`

With a plain-object registry and `backend in WEB_BACKEND_REGISTRY`, `getWebBackend("toString")` returns `Object.prototype.toString` — truthy, no `buildTurnTools` — and a core plugin contributing the id `constructor` false-collides at boot. PR2 feeds this from a nullable DB column, so a stale or hand-edited `provider.webBackend` becomes a TypeError mid-turn instead of the clean "no backend, no tools" degradation the ADR describes. `Object.create(null)`, or a `Map` with `.has()`, fixes it. `sandbox/index.ts` has the same latent flaw — I'd fix it here and follow up there separately rather than mirror it forward.

**A non-object entry crashes boot without attribution** — `plugins/loader.ts:~481`

`webBackends: [null]` throws `Cannot read properties of null (reading 'backend')`. Every other malformed case in this loop gets a plugin-named error, and "a third-party JS plugin can return anything" is the loop's whole justification, so the `typeof contribution !== "object" || contribution === null` guard completes the set.

**A malformed `content` payload reads as a successful empty page** — `web-backends/index.ts:367`

`typeof payload.content === "string" ? payload.content : ""` produces `{ content: "", truncated: false }`, which neither the model nor an Operator reading logs can tell from a genuinely empty page. The other narrowing paths warn; this one should too.

## Non-blocking

- **`plugins-list.tsx` wasn't updated.** `GET /plugins` now returns `webBackends`, but the admin list's `InstalledPlugin` type omits it and the emptiness test is still `toolSets.length === 0 && sandboxBackends.length === 0` — so a plugin contributing only a web backend renders "No contributions." If that's PR3 scope, worth saying so in the description, which currently reads as though the reporting path is complete.
- **`clearWebBackends()`** is a test-only export in a production module, where `sandbox/index.test.ts` resets via `vi.resetModules()`. I prefer the explicit reset, but it's the one place this isn't the mirror it's described as — worth a line saying why.
- **`readUrlInputSchema.max_length.max(1_000_000)`** (`types.ts:38`) silently duplicates `MAX_READ_URL_CONTENT_CHARS`; the import would be circular, so a cross-reference comment, or moving the constant into `types.ts`, keeps them from drifting.
- **Log volume:** the `dropped` warn and the scan warn are both per-call and model-triggerable, so a backend whose upstream mixes in unusable results warns on every search. `debug` for `dropped`, or warn-once per registration, might age better.
- **Double warn on an egress block:** the guard's own line plus `logCall(..., "blocked")`. Intentional per the comment (only the first carries the URL), but the outcome line adds nothing on that path.

Test coverage is genuinely good — keeping the real guard in the path with an injected resolver is the right call, and the `this`-preservation tests are the ones I'd have asked for. The gaps map to the findings: `createExecutors` throwing and hanging, URL and `content_type` length, a non-string `content`, and the scan-warn false positive.

MesoX:
Thanks — all six addressed, plus the four non-blocking notes. Three commits since
your review, since the branch was two behind when you read it:

- `15b319c` — the six findings, first pass
- `960577d` — `plugins-list.tsx`
- `5427ab8` — a second pass over the first two, prompted by re-reading your
  reasoning against what the fixes actually did (details below)

`pnpm typecheck`, `pnpm lint`, prettier and 1320 backend tests are clean.

## Blocking — `createExecutors` unguarded

`withTimeout(() => contribution.createExecutors(ctx, plugin), timeoutMs)` inside
a `try/catch` that warns and returns `{}`, so the factory degrades exactly like
the missing-`web_search` case two lines down. The arrow keeps it a method call on
the author's own object, so the `this`-preservation property holds. Two tests: a
factory that throws, and one that never settles.

Two things fell out of that which I've documented rather than changed, because
both are contract rather than code:

**`timeoutMs` now bounds the factory, and the SDK said otherwise.** The doc
comment read _"Per-call timeout for this backend's **executors**"_, and it
actively invites an author to tune the value to their upstream's search latency —
"a LAN metasearch should answer in ~2s". Guard the factory with that same value
and a backend whose `createExecutors` fetches a token, probes a health endpoint,
or warms a browser pool loses every web tool on every turn, with a warn line as
the only trace. `timeoutMs` now names both things it bounds and tells authors to
budget for the factory. `createExecutors` now documents the degradation contract
explicitly too, because boot is fail-loud for a _missing_ factory while runtime
is silent for a _throwing_ one, and an author would reasonably expect those to
match. Same note in the README's inline comment.

**The windows are additive.** `buildTurnTools` runs once per turn and each tool
call gets its own window, so the worst case a turn can spend inside a backend is
`(1 + calls) × timeoutMs` — 240s for a single search at the 120s boot ceiling.
Still bounded, which is the property the ceiling exists to protect, but no longer
one window, and the comment says so now. I considered a separate factory bound
and decided against it: two timeout knobs on one contribution is a worse contract
than one documented knob.

## Scan-bound warn

Gated on `results.length < MAX_SEARCH_RESULTS && entries.length > scanned.length`.
The `break` sits at the top of the loop, so a cap-filled result set means the
unscanned tail was never needed — the warn now fires only when a slot was still
open. Test: 500 usable results, no warn.

## `results[].url` and `content_type`

`MAX_URL_CHARS = 2048`, dropping the entry rather than truncating, since a cut
URL is a broken link. `content_type` shortened at 200 — but sliced **bare**
rather than through `truncate()`, which appends `…`. That marker is right for
prose a model reads and wrong for a machine-readable field, where a marked cut
leaves an invalid MIME type instead of an honestly shortened one.

Two adjustments beyond what you asked for, both because the first pass didn't
finish the job:

**Length is now checked before scheme.** `isPresentableUrl` parses the whole
string with `new URL()`, so the first version parsed a scan's worth of
multi-megabyte hrefs on the way to dropping them for length — the exact CPU
argument behind `MAX_SEARCH_RESULT_SCAN`, one operand order away. Same reorder on
the `read_url` path.

**The `read_url` fallback was unbounded.** When a backend's resolved URL is
rejected, `result.url` falls back to the model-supplied input — and
`readUrlInputSchema.url` was `z.string().url()` with no maximum, so that field
could still exceed the cap it had just gained. Taking your `types.ts` suggestion
closes both: `MAX_URL_CHARS` and `MAX_READ_URL_CONTENT_CHARS` now live in
`types.ts`, the schemas reference them directly instead of restating the numbers
behind a keep-in-sync comment, `url` is capped, and the `export *` at the bottom
of `index.ts` means every existing import still resolves.

## Registry prototype

`Object.create(null)`, with a test covering `toString`, `constructor` and
`__proto__`.

I left `sandbox/index.ts` alone here as you suggested — but that one turns out to
be **reachable, not latent**. `backend` is `z.string().min(1)` in the schema, and
`validateSandboxConfig` does `if (!registration) return null;` and then
`registration.configSchema.safeParse(...)`, so a sandbox saved with
`backend: "toString"` gets `Object.prototype.toString` back from the lookup and
500s on the next line. Admin-only, nothing leaks, but a real 500 rather than a
hypothetical one. Fixed in #388.

## Non-object contribution entry

Guard added, plugin-named, with a test.

Worth flagging that the `sandboxBackends` loop is missing **two** of these, not
one: no non-object guard, and no non-empty `backend` check either, so
`contributionId(undefined)` mints a plausible-looking `"acme.undefined"` — the
exact failure the webBackends loop's comment says its check exists to prevent.
Both in #388, along with `create` and the two schemas, which registered fine and
failed at chat-turn or save time.

## Malformed `content`

Warns now, matching every other narrowing path in the wrapper.

---

## Non-blocking

**`plugins-list.tsx`** — fixed in `960577d`, emptiness check included, so a
plugin contributing only a web backend renders its badge instead of "No
contributions."

**Docs site** — you were right that the description read as though the reporting
path were complete, and there's a second gap in it beyond `plugins-list.tsx`:
`apps/docs/content/extending/index.mdx` documents Sandbox backends and Tool sets
and has no web-backend section, while `packages/plugin-sdk/README.md` links to
that page as the full contribution reference. That one belongs with the docs work
in PR4 — the page describes what an Operator can actually use, and until the
Provider selector lands in PR3 there is nothing there to select — so shipping it
now would document an extension point nobody can reach. The SDK README, which
ships with the package an author builds against, _is_ updated here. The
description now states both halves explicitly rather than leaving "PR4 … + docs"
to carry it.

**`clearWebBackends()`** — the comment now states why it exists. Small correction
on the premise, though: `sandbox/index.test.ts` doesn't reset via
`vi.resetModules()`. Its comment reads _"The registry is module-level state; we
reset by re-importing via vi.resetModules isn't trivial here, so each test uses a
unique backend id"_ — a mangled sentence that scans as the opposite of what it
means. So the explicit reset here is a genuine divergence rather than a mirror of
an existing pattern, and the comment says that instead.

**`readUrlInputSchema.max_length.max(1_000_000)`** — no longer a duplicate; the
schema references `MAX_READ_URL_CONTENT_CHARS` directly, with a test asserting
the bound rather than a copied literal.

**Log volume** — agreed, `dropped` is now `debug`. It's per-call,
model-triggerable, and for an upstream that mixes in unpresentable results it's
expected steady state rather than a fault, so at `warn` a healthy backend logged
on every search a user ran. It also now carries `droppedScheme` and
`droppedLength` separately: an unusable scheme is a bug in the backend, an
over-length URL is noise, and a single total couldn't tell an Operator which they
had. The scan warn and the missing-content warn stay at `warn` — both mean the
backend is misbehaving, not merely noisy.

**Double warn on an egress block** — I'd keep this one. The second line isn't
redundant so much as differently shaped: `logCall` emits the uniform
`{ backend, plugin, tool, durationMs, outcome }` record for every executor call,
and it's the line log-based alerting keys on. Dropping it for the blocked path
would make `outcome: "blocked"` the one outcome missing from that series, and the
duration would go with it. The guard's line carries the URL and the reason; this
one carries the shape.

willdady (approved):
Approving — this is in good shape and I'm merging it as-is.

The details I want to call out as correct, because they're the ones that would have
been easy to get wrong: passing the contribution by reference instead of spreading
it (so a class-instance contribution keeps its prototype and `this`), validating
`backend`/`name` before namespacing rather than after, and reading executor
payloads as `unknown` with field-by-field narrowing. The third one in particular
is the difference between "the SDK types describe this" and "a third-party JS
plugin cannot break core," and the `title` → `""` case is the right call over
`String()` coercion. Test coverage asserting against the constants rather than
hand-copied numbers is also appreciated.

**On your `isPresentableUrl` question:** leave it in `web-backends/` for now, lift
it into `@platypus/schemas` in PR3 when the frontend actually becomes the second
consumer. One consumer doesn't justify the shared-package indirection yet, and by
PR3 you'll know whether the frontend wants exactly this predicate or something
that also handles the rendering-time concerns. No need to change anything here.

**One thing worth picking up in PR2** — the `MAX_URL_CHARS` cap on
`readUrlInputSchema.url`. The invariant you're protecting is real (the input is
the fallback when a backend's resolved URL is rejected, so an uncapped input would
let that field exceed the cap), but bounding the _input_ to get there has a
user-visible cost once the feature is reachable: presigned S3 URLs, Azure SAS
links, and some SharePoint/OAuth URLs legitimately run past 2048 chars, and those
now fail at the schema as an AI-SDK input-validation error rather than through the
graceful `{ error }` contract the rest of the module maintains. It's also the one
place `read_url` diverges from `fetchUrl`'s input, which is otherwise mirrored
exactly.

You can keep the invariant by bounding the fallback instead of the input:
`payload.url` when presentable and within the cap, else the requested `url` when
it's within the cap, else `new URL(url).origin`. Cheapest to change now while
nothing depends on the schema shape, so PR2 rather than a follow-up.

Three smaller non-blocking notes, all inside the new module and none with a
current caller, so entirely up to you whether they ride along with PR2:

- `droppedLength` counts `typeof entry.url !== "string"` alongside genuinely
  over-length URLs. The comment justifying two counters puts a missing field in
  the "backend bug" category, but it lands in the "routine noise" bucket, so a
  broken backend reads as harmless. A third counter would keep the field meaning
  what the comment says.
- The `MAX_READ_URL_CONTENT_CHARS` comment says the cap bounds context, but
  `max_length` maxes at the same value, so one call can still return ~1M chars.
  Inherited from `fetchUrl` and not a regression — but core owns this schema
  outright, unlike `fetchUrl`'s, so the two levers don't have to share a number.
- A blocked egress emits two `warn` lines in different shapes (the
  `fetchUrl`-matching line plus the outcome line). Both have a rationale; folding
  `reason` into the outcome line would give one structured record.

For PR4: the constants ADR-0014 is silent on — `MAX_ANSWER_CHARS`,
`MAX_SEARCH_RESULT_SCAN`, `MAX_URL_CHARS`, `MAX_CONTENT_TYPE_CHARS`,
`MAX_READ_URL_CONTENT_CHARS`, and the `isPresentableUrl` scheme filter — are
core-owned enforcement decisions rather than implementation detail. Since 0014 is
still `proposed`, amending its Consequences is fair game before you flip it to
`accepted`; that's the right home for the table in this PR description.

Looking forward to PR2.

PR #388, opened 2026-07-29
fix(backend): harden the sandbox registry and contribution validation

MesoX:
Follow-up to the review on #384, which flagged the `Object.prototype` registry
flaw as shared with `sandbox/index.ts` and asked for it to be fixed separately
rather than mirrored forward. Reviewing that seam turned up the same class of
problem in the loader, and one path that is reachable today.

Based on `main`, independent of #384 — nothing here touches the web-search
backend.

## The 500

`getSandboxBackend` was backed by `{}`. `backend` reaches it from two sources
neither of which is a closed set — a request body, where the schema is
`backend: z.string().min(1)`, and the `sandbox.backend` column — and both
`validateSandboxConfig` and `validateSandboxCredentials` are written to treat an
unregistered id as "nothing to validate":

```ts
const registration = getSandboxBackend(backend);
if (!registration) return null;
const result = registration.configSchema.safeParse(config ?? {});
```

`POST /organizations/:orgId/workspaces/:workspaceId/sandboxes` with
`backend: "toString"` returns `Object.prototype.toString` from that lookup —
truthy, with no `configSchema` — so the next line is `undefined.safeParse(...)`:
a TypeError surfacing as a 500, rather than the clean pass-through the guard
intends. `"constructor"`, `"hasOwnProperty"` and friends do the same.

Workspace-config admin is required and nothing leaks, so this is a robustness
fix rather than a security one. `Object.create(null)` closes it: a null-prototype
registry has nothing to inherit, so an unknown id is `undefined` whatever it is
called.

## The boot validation

The `sandboxBackends` loop went straight to
`contributionId(contribution.backend)`. Every shape below is guaranteed by
`SandboxBackendContribution` and none of it is guaranteed by a third-party _JS_
plugin — which is the loop's whole justification — and each one surfaced far from
its cause:

| Malformed contribution                  | Before                                                                | Now                          |
| --------------------------------------- | --------------------------------------------------------------------- | ---------------------------- |
| `sandboxBackends: [null]`               | `Cannot read properties of null (reading 'backend')`, no plugin named | boot error naming the plugin |
| no `backend`                            | `contributionId(undefined)` registers `"acme.undefined"`              | boot error                   |
| no `name`                               | blank entry in `GET /sandbox/backends`                                | boot error                   |
| no `create`                             | TypeError at chat-turn time                                           | boot error                   |
| no `configSchema` / `credentialsSchema` | 500 on the Operator's sandbox form                                    | boot error                   |

This is the set the `webBackends` loop in #384 already enforces, so the two loops
now read the same. The schema check runs after the factory form is resolved and
is duck-typed on `safeParse` rather than `instanceof z.ZodType`, because that is
exactly what the three static consumers call.

**Compatibility note:** these are new boot-time rejections. A third-party sandbox
plugin that omitted a field TypeScript already requires, and that happened not to
hit the failing path yet, will now fail to boot with its name in the message.
That is the ADR-0013 fail-loud posture, and it is why this is a separate PR
rather than folded into #384. All three in-repo contributions (`docker`, `ssh`,
`example`) supply every field.

## Tests

- `sandbox/index.test.ts` — `toString`, `constructor`, `__proto__` and
  `hasOwnProperty` all resolve to `undefined`
- `plugins/loader.test.ts` — one table-driven case per malformed shape, each
  asserting the plugin name is in the message

1271 backend tests pass; typecheck, lint and prettier clean.

### DIFF
--- a/apps/backend/src/plugins/loader.test.ts
+++ b/apps/backend/src/plugins/loader.test.ts
@@ -522,6 +522,97 @@ describe("loadPlugins — sandbox backends", () => {
     ).rejects.toThrow(/@bad\/plugin.*sandboxBackends.*array/s);
   });
 
+  // `sandboxBackends` being an array is checked above; a third-party JS plugin can
+  // still put anything *inside* it. Each of these previously surfaced far from its
+  // cause — an unattributed TypeError at boot, a mystery `"acme.undefined"` in the
+  // catalog, or a 500 on an Operator's sandbox form.
+  it.each([
+    [
+      "a non-object entry",
+      null,
+      /@platypus\/bad.*sandbox backend contribution must be an object/s,
+    ],
+    [
+      "a missing backend id",
+      {
+        name: "x",
+        configSchema: z.object({}),
+        credentialsSchema: z.object({}),
+        create: () => ({}),
+      },
+      /non-empty "backend" id/s,
+    ],
+    [
+      "a missing name",
+      {
+        backend: "b",
+        configSchema: z.object({}),
+        credentialsSchema: z.object({}),
+        create: () => ({}),
+      },
+      /must declare a non-empty "name"/s,
+    ],
+    [
+      "a missing create",
+      {
+        backend: "b",
+        name: "x",
+        configSchema: z.object({}),
+        credentialsSchema: z.object({}),
+      },
+      /must provide a "create" function/s,
+    ],
+    [
+      "a missing configSchema",
+      {
+        backend: "b",
+        name: "x",
+        credentialsSchema: z.object({}),
+        create: () => ({}),
+      },
+      /must provide a Zod "configSchema"/s,
+    ],
+    [
+      "a missing credentialsSchema",
+      {
+        backend: "b",
+        name: "x",
+        configSchema: z.object({}),
+        create: () => ({}),
+      },
+      /must provide a Zod "credentialsSchema"/s,
+    ],
+  ])(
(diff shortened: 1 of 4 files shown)


willdady (approved):
Approving. Verified against the PR head in an isolated worktree: full backend suite 1327/1327, `tsc --noEmit` clean, Prettier clean, docs-contract 22/22. I also reverted just the two source files and confirmed all 7 new tests go red against `main` — they are genuine regression tests.

Both changes land at the right layer: `Object.create(null)` fixes the 500 at the registry rather than at each of the four `getSandboxBackend` call sites, and it matches the form `web-backends/index.ts:66` already uses. The loader checks reject nothing that satisfies the published `SandboxBackendContribution` type, so the compatibility note is accurate.

Merging as-is. Three follow-ups going out in a separate PR rather than blocking this one:

1. A throwing `configSchema` **factory** is still unattributed — `loader.ts:475` calls it bare, one line above the new check. Confirmed empirically: a factory doing `cfg.region.toUpperCase()` on absent config aborts boot with `Cannot read properties of undefined`, no plugin named. The new guard catches a factory that *returns* junk, not one that *throws*.
2. Five of the six new test regexes don't pin the plugin name, though the description says they do — attribution could be dropped from those messages and the suite would stay green.
3. `extending/index.mdx` gets a clause on the new boot rejections, per the CLAUDE.md docs table.

Also filing the last instance of the prototype class: `TOOL_SETS_REGISTRY` (`tools/index.ts:26`) is still a plain `{}`.

PR #393, opened 2026-07-30
feat: serve chat search from a Web-search backend

MesoX:
Wires the Extension point from #384 into a turn: a Provider now stores the Web-search backend it serves search with, and a configured backend takes precedence over the Provider's native tool. Part 2 of 4 for #329, on [ADR-0014](docs/adr/0014-web-search-backend-extension-point.md).

Two kinds of change are separated below, because they are not the same thing to review: what #384's **approval review** asked to pick up here, and what a **review pass on this branch** turned up afterwards. Nothing from #384's earlier `CHANGES_REQUESTED` review reappears — those six findings were fixed inside #384 before it merged.

## What's here

- **`providerHasNativeSearch`** in `@platypus/schemas` — a positive capability predicate over `providerType` + `apiMode`, replacing the frontend's hardcoded `!== "Bedrock"` exclusion. It lives beside `defaultPassthroughFileTypes` for the same reason: the backend's injection gate and PR3's toggle-visibility gate must agree, and mirrored copies of a capability table drift. Verified one-for-one against `openProvider`'s actual `searchTools()` map — Anthropic, Google and OpenRouter yes; OpenAI on the Responses API yes and on chat no; Bedrock and anything unknown no.
- **`provider.webBackend`** — nullable text, migration `0050`. Free text, not an enum or FK: the valid set is whichever plugins the deployment loaded, which the database cannot know. No backfill. Bounded at 200 chars, and `""` normalises to null.
- **Turn-time resolution** — `resolveSearchMode(request.search, provider)` returns `none | native | { backend }`, and `resolveSearchTools` builds the tools. Explicit-plugin-first: a configured backend wins over native search, and a stored id whose plugin has since been removed from `PLATYPUS_PLUGINS` degrades to no search tools plus a warn — never a silent fallback to native, which would serve a different search than the Operator selected.

Resolution is awaited **inside** the existing `Promise.all` alongside `loadTools` and the sub-agent load, not after it. A backend's `createExecutors` may open a connection or warm a pool, and its `timeoutMs` budget (30s default, 120s ceiling) would otherwise land on first-token latency on top of the other network waits instead of beside them. The assignment order that gives search its precedence is unchanged: loaded tools, then search, then sub-agent tools.

## Two behaviour changes worth flagging

Neither was raised in review; both fall out of implementing ADR-0014's capability gate, and both are Operator-visible, so they are called out rather than left to be discovered.

**This adds backend-side capability gating that did not exist before.** The old gate was `requestedSearch && nativeSearchEnabled !== false`, with the Provider-capability check living only in the frontend. A stale client sending `search: true` at a no-native-search Provider with no backend configured now gets no tools instead of the dead OpenAI-hosted `web_search`. That is ADR-0014's stale-client case, not a regression — that client got an empty-in-practice tool anyway.

**`nativeSearchEnabled: false` disables plugin search too.** Deliberate: the switch currently means "no search on this Provider at all", and re-cutting those semantics is deferred. The consequence is worth stating plainly, because it is the opposite of intuitive for the Providers this feature exists for: an Operator who switched the toggle off on a vLLM Provider _because native search never worked there_ will select a Web-search backend and find it silently dead. The field name and the form label both say "native", which stopped being accurate here. PR3 carries the user-visible half — relabel the switch to web search, helper text stating it governs both paths, and the selector reading as inert while the switch is off. The rename itself stays deferred.

## Asked for in #384's approval review

**The one explicit ask — `readUrlInputSchema.url`.** Landed as its own commit (`c98815f`) so it reads separately from the new wiring. `.max(MAX_URL_CHARS)` is gone from the input; the returned field is bounded down every branch instead — `payload.url` when presentable and within the cap, else the requested `url` when within it, else `new URL(url).origin`, which is the shape #384's review prescribed. Three branches rather than two precisely because the input is now uncapped, so the request is no longer itself a bounded fallback. Tests: an over-length presigned URL reaches the executor whole and comes back as its origin, and a short request still wins over an over-length resolved URL.

**All three optional notes from that review, taken:**

| Note on #384                                                                                     | Shipped here                                                                                                                          |
| ------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------- |
| `droppedLength` folds in a non-string `url`, so a broken backend reads as routine noise          | A third counter, `droppedNoUrl`, alongside `droppedLength` and `droppedScheme`.                                                       |
| `MAX_READ_URL_CONTENT_CHARS` is called a context bound, but `max_length` maxes at the same value | Split into two levers: the content cap bounds a whole paginated read, `MAX_READ_URL_SLICE_CHARS = 100_000` bounds any one page of it. |
| A blocked egress emits two `warn` lines in different shapes                                      | One record — `url` and `reason` ride the uniform `{ backend, plugin, tool, durationMs, outcome }` line.                               |

On that last one: it **reverses the position argued in my comment on #384** ("the second line isn't redundant so much as differently shaped — it's the line log-based alerting keys on"). The approval review re-raised the fold after that comment, which makes it the standing preference, so it is done. Flagging the reversal rather than letting it read as silent drift.

**`isPresentableUrl` stayed in `web-backends/`,** per the same review's answer. PR3 will decide whether the frontend wants exactly this predicate or one that also handles rendering-time concerns.

## Found in a review pass on this branch

Not raised by anyone on #384 — these came out of re-reading the three commits above against ADR-0014 and against #384's own reasoning. Two of them are corrections to the optional notes: the first attempt implemented each ask and overshot it.

- **`max_length` is clamped in `execute`, not bounded by the schema** (`423917b`). Splitting the two levers by lowering the schema bound to 100 000 reintroduced exactly the failure mode the `url` cap was reverted for one commit earlier: a model that learned `fetchUrl`'s `1_000_000` ceiling would fail as an AI-SDK input-validation error rather than through this module's `{ error }` contract. Clamping needs no error path — the page comes back shorter and `next_start_index` already tells the model the read continues. The input mirrors `fetchUrl` byte-for-byte again.
- **The blocked-egress record kept its shape but lost the event's name** (`423917b`). After folding, a block logged "Web backend executor call failed" — where nothing failed and the executor was never called — while `fetchUrl` logs "Blocked a model-supplied URL by network policy" for the same policy decision. The message is now outcome-derived, so one log query spans both tools that fetch a model-supplied URL, and `durationMs` is documented as time-in-call-so-far: on a block it measures the egress guard's DNS resolution.
- **The search decision was made twice, and one branch could not be false** (`1a74bf6`). The first cut was a boolean `shouldServeSearch` plus a re-derived branch at the injection site, which left `else if (providerHasNativeSearch(provider))` unreachable-false — once the gate passes with a falsy `webBackend`, native capability is already implied, so a reader had to prove that implication to know search could not silently vanish. `resolveSearchMode` returns `none | native | { backend }` instead, carrying the id with the decision.
- **The backend factory sat on the critical path** (`1a74bf6`). It was awaited after the existing `Promise.all`, so `createExecutors` and its `timeoutMs` budget added to first-token latency instead of overlapping the tool and sub-agent loads. Now inside the `Promise.all`; precedence unchanged.
- **The stale-backend warn could not identify the row to fix** (`1a74bf6`). It named `workspaceId`, but an org-scoped Shared Provider is one row serving many Workspaces (ADR-0007), so the same stale id warns under many Workspaces while the remedy is a Provider edit. It now carries `orgId` and `providerId`.
- **The content-cap warn had no plugin attribution** (`423917b`) — the only line in the module without it, while `pluginName` documents itself as being there for attribution in every line. Pre-existing in #384, adjacent to code this branch was already touching.
- **`webBackend` was unbounded free text** (`1a74bf6`). It reaches a log line on every searching turn, so it is now `max(200)`; and `""` normalises to null, because a form select's "None" yields the empty string and would otherwise be a second representation of "no backend" in a nullable column. Normalised rather than rejected — `.min(1)` would turn choosing None into a 400.
- **A `javascript:` _requested_ URL was unpinned** (`423917b`). `z.string().url()` accepts it, so the egress guard's scheme check is the only thing keeping it out of `result.url`, which PR3 renders as a clickable pill. Now covered by a test rather than left as an implicit coupling.
- **The `""` → null normalisation made `webBackend` a required key** (`a1f8a41`, caught by this PR's own CI run). `.optional().transform(...)` wraps the field in `ZodEffects`, which hides the optionality `z.infer` reads off the shape — so `Provider` gained a mandatory `webBackend` whose type merely included `undefined`, and every Provider literal omitting the field stopped compiling. `.optional()` now sits outside the transform, with a comment, since the two spellings look interchangeable and are not.

## Verification

- `pnpm typecheck`, `pnpm lint` and prettier clean. Backend **1345**, frontend 26, schemas **73**, plugin-sdk 7, example-plugin 3.
- New tests: the whole `providerHasNativeSearch` table (5 types × both API modes, plus an unknown type); `resolveSearchMode` across toggle-off, switch-off ± backend, Bedrock ± backend, OpenAI-chat ± backend, and `""` reading as no backend; resolution asserting the native tool is neither injected nor even constructed when a backend is set; a `web_search`-only backend injecting only search; a since-removed id serving nothing with the warn's fields pinned; `webBackend` round-tripping create/update including the `""` → null and 200-char boundaries.
- **Trigger-initiated runs are covered by a test, not by inspection**: `runMode: "headless"` — what `AgentRunner` passes for a non-user principal — with search on and a backend configured serves the backend's tools. There is one injection site for every run path, and the test pins that it is not mode-gated.
- `read_url`: a `fetchUrl`-sized `max_length` clamps to one page with a continuation index rather than erroring; the tail of a capped page still emits the terminal line with no `next_start_index`; the blocked-egress path emits exactly one warn, with the message pinned.

## Docs — deliberately PR4, and checked rather than assumed

I ran this diff against the "ship docs in the same PR" table in `CLAUDE.md`: no `.env.example`, nothing under `apps/backend/src/plugins/**`, no visible frontend label. The one row that does fire is the user-facing `max` in `packages/schemas` — the new 200-char bound on `webBackend`.

That is left to PR4 with the specific values written down, on the same argument accepted for #384's docs deferral: until PR3's selector lands there is no way for an Operator to set `webBackend` at all, so documenting a limit on it describes an extension point nobody can reach. PR4 is a planned PR in this stack rather than the un-filed follow-up the rule exists to prevent, and it is where the `extending` page gains its web-backend section in the first place. Say the word if you would rather it ride here instead.

## For PR4

The ask on #384 to lift the core-owned constants into ADR-0014's Consequences before the flip now has five additions from this PR: `MAX_READ_URL_SLICE_CHARS` and the fact it is clamped rather than schema-enforced; an exception clause on the observability contract, which said URLs are "never on the outcome line" and now must record that a block puts `url` and `reason` there, along with what `durationMs` means on that path; `nativeSearchEnabled: false` disabling plugin search; the 200-char `webBackend` bound; and one stated limitation — **there is no teardown hook**. `createExecutors` runs once per turn and whatever it opens is dropped when the turn ends. That matches sandbox adapters, which are instantiated per turn and discarded the same way, and ADR-0013's in-process posture, so adding a `dispose` belongs to one PR covering both Extension points rather than making web backends the only shape with a lifecycle. Flagging rather than fixing it — tell me if you would rather see it sooner.

## Follows

PR3 catalog route, Provider selector, `canSearch` rework, the switch relabel, Sources rendering · PR4 ADR → `accepted` + docs.

Refs #329 — part 2 of 4, so this does not close it.


willdady:
Reviewed the full diff. Verdict: **merge**, with one question below and the rest tracked into PR3/PR4.

Checks that hold up, so nobody re-does them: `buildTurnTools` already wraps `createExecutors` in `withTimeout` + try/catch and returns `{}`, so moving resolution inside the `Promise.all` introduces no rejection path for the turn. The capability table matches `openProvider`'s `searchTools` map one-for-one. `presentableReadUrl`'s "unreachable" comment is accurate — `egress-guard.ts` gates on `ALLOWED_PROTOCOLS = {http:, https:}`, so an echoed request URL is always http(s). And `.optional()` outside the transform is load-bearing beyond the type error it was made for: it keeps the key genuinely absent from the parse result, so `provider.ts`'s `.set({...data})` won't null the column when the current form saves without `webBackend`.

### Before merge (one question)

- [ ] **Has #384 shipped in a release?** The blocked-egress log message changes here from `"Web backend executor call failed"` to `"Blocked a model-supplied URL by network policy"`. If #384 is already out, that's an Operator-visible rename on a string alert rules key on. If it isn't, there's nothing to decide.

### PR3

- [ ] **Replace the frontend capability check** — `apps/frontend/components/chat.tsx:413` still hardcodes `providerType !== "Bedrock"`, so this PR creates the second copy of the table rather than collapsing it. Already scoped to PR3's `canSearch` rework; the only real constraint is that PR2 and PR3 land in the **same release**. If PR2 ships alone, an OpenAI provider on `apiMode: "chat"` (vLLM, llama.cpp) with the switch on still renders the search toggle while the backend now serves zero tools for it — the toggle goes from broken-loudly to silently inert, which is the worse one to diagnose.
- [ ] **Loosen `providerHasNativeSearch`'s signature** to match `defaultPassthroughFileTypes` (`{providerType: string; apiMode?: string}`). The sibling's loose typing is deliberate and commented — so both sides can pass their own provider shape. `Pick<Provider, ...>` also makes the `default:` branch unreachable per TS while its doc comment says it exists for unknown types, which is the runtime case a text column really has. Widening a parameter is backwards-compatible, so this can happen any time; PR3 is the right moment only because that's when the frontend call site appears and would otherwise carry a cast that outlives its reason.

### PR4

- [ ] **The 200-char `webBackend` bound needs documenting.** Deferral accepted on the stated argument — nothing can set the field until PR3's selector lands. Recording it here because I checked and _nothing mechanical catches this_: `docs-contract.test.ts` pins field limits only via its explicit claim list, so CI will stay green with the bound undocumented. That makes it the exact shape of the miss `CLAUDE.md` cites `627cb1e` for, and the deferral is only sound while PR4 doesn't slip.
- [ ] **`max_length`'s `.describe()` over-promises** — the schema advertises up to 1,000,000 while `execute` clamps to `MAX_READ_URL_SLICE_CHARS`. Clamp-not-reject is the right call and `next_start_index` carries the continuation; stating the effective per-call page size in the description just saves a model inferring the clamp from a short page.
- [ ] Optional, while the file is open: two of the new comment blocks record PR history rather than current behaviour ("A blocked egress used to emit this line _and_ a second…", "The two used to share `MAX_READ_URL_CONTENT_CHARS`"). The rationale is worth keeping; present tense keeps the same information without dating it.

Nothing above touches migration `0050` or a persisted shape, which is why none of it blocks. The resolution model is a real improvement over the boolean, and the test that asserts the native tool is never _constructed_ — not merely absent from the output — is the right shape for a precedence claim.

willdady:
Resolved my own pre-merge question, so no need to spend time on it: **#384 has not shipped in a release.** It merged 2026-07-29 (`6c68205`), the latest release `v2.2.1` was cut 2026-07-26, and that commit is not an ancestor of the tag. The old `"Web backend executor call failed"` string has therefore never reached an Operator, so the rename to `"Blocked a model-supplied URL by network policy"` costs nothing and there is no second rename to worry about.

That was the only item gating merge. The PR3 and PR4 checkboxes above stand as tracked follow-ups rather than changes requested here.

willdady (approved):
Approving on the review above: checks green, nothing blocking. The remaining items are tracked as PR3/PR4 follow-ups, and the one pre-merge question resolved itself — #384 is unreleased, so the log rename costs nothing.

Merging on the understanding that PR3 lands in the same release, so the frontend's `providerType !== "Bedrock"` check doesn't ship against the new backend-side capability gate.

PR #412, opened 2026-08-04
feat: select a Web-search backend on the Provider form

MesoX:
# PR3 — Catalog route, Provider-form selector, `canSearch`, Sources rendering

Closes the operator-facing half of #329. Third in the stack behind #384 (PR1) and #393 (PR2). Branch `feature/web-backend-catalog-and-selector`, now merged up to `main` @ `41f69b5` (2.5.0).

Until now a Web-search backend could only be selected by `PUT`ing `webBackend` on the Provider API, and on a Provider with no native search the Chat **Search** toggle stayed hidden — so the endpoints this Extension point exists for (vLLM, LiteLLM, Bedrock) could not reach a backend from Chat at all. This makes the selection reachable and presents a backend's results the way native search's are.

**20 files, +1586/−83.** No migration, no schema change, no change to what a searching turn resolves to.

> **Opened as a draft on purpose**, to settle the open questions at the bottom and the fact that the control this PR adds is collapsed by the next one (PR3.5, below) before anyone spent a full review pass on it. The 2026-08-08 review answered all five and asked for four changes; those are in as `739807f` and `3f5e086`, and the thread carries the response.

## What's in it

**`GET /organizations/:orgId/web-backends`** — `{ results: { backend, name, plugin }[] }` from `getWebBackends()` + `getWebBackendPlugin`.

Mounted **org-scoped**, which is a deliberate correction to "mirror sandbox's `GET /backends`". A sandbox _is_ a workspace resource; a Provider is not — Providers exist in both scopes (ADR-0007) and one `ProviderForm` serves both from an `orgId` plus an _optional_ `workspaceId`. A workspace-scoped catalog is unreachable exactly when an Org Admin edits a Shared Provider, which is the case the selector matters most for. The precedent is `GET /plugins`, mounted the same way and commented for the same reason: the orgId scopes _access_, not the data.

No scope filtering is possible or needed: a backend arrives via `PLATYPUS_PLUGINS` in the backend environment and is registered at boot into a module-level registry. There is no table behind it, so no org or workspace can hold a different list.

**Provider form** — a nullable select under Advanced settings, beside the Native web search switch, fed by the route through SWR. Shown for every Provider type, including ones with native search, since a set backend wins over native by design.

**`canSearch`** — now `nativeSearchEnabled !== false && (providerHasNativeSearch(p) || Boolean(p.webBackend))`, which is ADR-0014's gate formula verbatim. Gated on the stored id alone: no catalog fetch in `chat.tsx`, no liveness `some()` (that is PR9). A stale id degrades server-side to no tools plus a warn line.

**Sources rendering** — native search surfaces citations as `source-url` parts. A plugin `web_search` is client-executed, so without lifting its result the same toggle gives pills on Anthropic and nothing on vLLM. Its `results[].url` / `title` now merge into the one `<Sources>` row, deduplicated by URL.

## Five decisions worth your attention

**1. The `web_search` tool card is compact, not the generic renderer.** Adding pills alone would have given a vLLM user pills _and_ a collapsible raw-JSON block repeating every result, where Anthropic shows pills alone — which is the outcome ADR-0014's "Presenting results" section rejects. So `tool-web_search` gets its own branch beside `LoadSkillTool`: header with the query, body with the answer box and a result count, no JSON. It stays rather than being hidden entirely because the pills cannot say that a search ran, what was searched for, or that it failed — and a failed search returns `{ error }` on a successful part, so there is nothing in the Sources row to notice. `read_url` keeps the generic card; it is a fetch, not a citation source.

**2. The select is not disabled while Native web search is off.** The plan's wording was "reads as disabled/inert", but a genuinely `disabled` select traps a stored backend: clearing it would mean switching native search on, clearing, and switching off again. Concealing a stored value is the same category of trap this section exists to close, so the control stays interactive and the helper text turns into a visible warning instead. Same reasoning behind showing a stored id the catalog no longer lists, labelled `(not installed)`, rather than hiding the field on a deployment with no backends when a value is set.

That visibility rule had one more hole, closed in the last two commits: it reads the value the field itself edits, so on a deployment with nothing installed, clearing a stale id — choosing None, the recovery path — emptied `webBackend`, turned the condition false and unmounted the control mid-interaction. No confirmation and no undo short of a reload. The field is now latched open from the selector's own handler once the reader has used it. Set from the handler rather than derived, so the latch cannot hold the field open for a reason the reader never caused.

**3. `isPresentableUrl` moved to `@platypus/schemas` — which reverses the answer you gave on #384's approval** ("stays in `web-backends/`"). That answer deferred the decision to this PR, on the condition that the frontend becomes the second consumer, which it now is; recording the reversal rather than letting it read as drift, as with PR2's two-warn fold. The frontend needs exactly the scheme filter — length is bounded upstream by `MAX_URL_CHARS`, with the opposite treatment (drop, not truncate) — so it is lifted verbatim rather than mirrored. Two copies of a security predicate drift, and the frontend's copy is the one deciding what becomes an `href`. Nothing in `apps/frontend` imports backend runtime code, so "leave core's alone" would have meant a mirrored predicate, not a shared one. The re-export from `web-backends/` is gone and every caller imports from the package; the cases live in `packages/schemas/index.test.ts`.

**4. Telling a plugin `web_search` from the Provider's own.** They share a tool name: `services/provider.ts` registers native search under `web_search` on OpenAI, OpenRouter and Anthropic. Matching the name alone — the first shape of this PR — sent a native search to the compact card, which claimed "0 results" for a search that returned ten and hid the vendor payload. One `isPluginWebSearchPart` predicate now gates both the card and the Sources lifting, layered by how much each check knows:

1. `providerExecuted === true` ⇒ native. Definitive where a provider sets it; Anthropic and OpenAI both do, from the first chunk, so there is no flicker.
2. A **core-stamped marker** ⇒ plugin. `composeWebBackend` sets `metadata: { platypusWebBackend: true }` on both Tools it builds, and the AI SDK carries a Tool's `metadata` onto the tool call's `toolMetadata`, onto the UI part, and into the stored message. This is the only check that does not depend on a vendor reporting itself — and it is needed, because **`@openrouter/ai-sdk-provider` never sets `providerExecuted` anywhere in the package**. On `providerExecuted` alone, a native OpenRouter search would have kept the original bug, in a worse form: no output to count, so a permanent "Searching…".
3. Core's own result shape, for a finished call with no marker — a message stored before the marker existed. Vendor payloads are shaped differently and do not match.
4. Nothing else. This step first read a bare `input-streaming` as plugin, on the belief that a marker is attached only once the tool call is parsed. That is wrong: `ai@7.0.48` copies a Tool's `metadata` onto the `tool-input-start` chunk and the UI stream forwards it onto the part, so one of our own calls is **already marked while streaming** and check 2 has it. The only parts reaching step 4 were the unmarked native ones it existed to exclude — so a native OpenRouter search rendered as the compact card mid-stream and swapped to the generic renderer at `input-available`. Fixed in the last two commits, with a test for a streaming part on each side of the line.

Anything left over reads as native and keeps the generic renderer, which shows whatever it really is. **I could not verify OpenRouter empirically — no account** — so the fix is written not to need the answer: the marker is true of core's tools on every provider, verified or not. The marker key is exported from `@platypus/schemas` for decision 3's reason, and pinned by a backend test on the side that writes it.

This is a dependency ADR-0014 does not record. Its Citations consequence says the frontend reads the core-owned `results[].url`; it must also decide _which_ `web_search` a part is before reading that shape at all. Queued as a sixth Consequences amendment for PR4, together with the cost it names: the deliberate snake_case exception buys prompt portability by sharing native search's tool name, and the price is that core cannot be identified by name.

Queued with it, from the correction above: the discriminator depends on the SDK carrying that metadata **from the first streaming chunk**, not only at parse time. Worth writing down as the version-pinned assumption it is — though not worth a test asserting a third party's stream shape, because the failure mode is cosmetic: if a future SDK stopped carrying it there, a plugin search would render generic mid-stream and correct itself at `input-available`. What is pinned is the side we own, that `composeWebBackend` stamps it.

**5. A native citation pill is now titled, and scheme-checked.** A `source-url` part carries an optional `title` and Anthropic sends one, so native Providers were showing raw URLs in the same row where a backend Provider shows real titles. The pill is now titled by it, falling back to the URL — `||` and not `??`, since a vendor sending an empty title would otherwise render a pill with no label.

That change is what pulls native parts through `isPresentableUrl` as well, which the plan scoped to backend-supplied URLs only. Not distrust of the vendor: with the URL off the screen, a `javascript:` or `data:` citation would render as an ordinary-looking link and the raw-URL label was the only thing that made one visible. **Stated rather than implied, since it is a behaviour change on a shipped native path:** an unpresentable native citation is now dropped from the row and from its count, not degraded to unlinked text. No live loss — Anthropic and OpenAI emit `http(s)` only — but it is core's predicate deciding what a vendor's citation renders as, so it joins amendment 6 rather than passing as implementation detail.

The render-side check is not redundant with the backend-side drop. The egress guard covers **model-supplied** URLs entering `read_url`; this covers **backend-supplied** URLs becoming clickable, which nothing else covers, and a backend is third-party code. `javascript:` and `data:` results are dropped, and they are dropped before both counts — the Sources row's and the card's, which now counts through the same helper the row is built from rather than off the raw array. A "3 results — listed above as sources" line above one pill is its own bug. The card counts **its own** call deliberately: two searches in one turn that both surface a page total one more than the row, which deduplicates message-wide, and the alternative is a card reporting a number with nothing to do with the search it heads.

## Exercised against a real backend

Core ships no backend by design (ADR-0014 non-goal), which makes this stack hard to try. So I wrote one, out of tree and public, to have something to point a Provider at: **https://github.com/MesoX/platypus-searx-backend**

It is a plain `webBackends` contribution — `web_search` over a self-hosted [SearXNG](https://docs.searxng.org/), `read_url` over [obscura](https://github.com/h4ckf0r0day/obscura) via CDP — so **no search API key and no third-party account** are needed to exercise the Extension point end to end. `read_url` is contributed only when a browser URL is configured, which also makes the optional-second-tool path a real case rather than a test fixture. The README carries the compose snippets and the two non-default SearXNG settings it needs.

I am not proposing it as a dependency or asking for it to be adopted — it is a reference implementation and a test harness. It happens to be the first third-party consumer of the contract shipped in #384/#393, which is what makes it useful here: everything in this section was found by _using_ the seam, not by reading it.

**It works.** A self-hosted deployment on a test box, Provider `Local` (OpenAI + `apiMode: "chat"`, the vLLM shape), backend `searx.web`: search, citations, paginated reading, the egress block and the stale-selection degrade all behave as specified. The scenario-by-scenario account, with the signal that distinguishes each one and an explicit list of what the manual pass did **not** cover, is under **Verification → Real-world verification** below.

Two things came out of running it that a code review will not surface, below.

## PR3.5 is a real dependency of this one, not a nice-to-have

This PR ships **two controls fighting over one slot** — a `nativeSearchEnabled` switch and a `webBackend` select, where the switch is a master off for both. That arrangement has no good version, and the state table shows why:

| `nativeSearchEnabled` | `webBackend` | Chat globe | A searching turn runs             |
| --------------------- | ------------ | ---------- | --------------------------------- |
| on                    | None         | shown      | native vendor search              |
| on                    | set          | shown      | the backend; native never reached |
| off                   | None         | hidden     | nothing                           |
| off                   | set          | hidden     | nothing                           |

Row 4 is the trap: an Operator who switched native search off on a vLLM Provider _because native search never worked there_ selects a backend and finds it silently dead. Flipping row 4 the other way does not fix it — it makes the switch native-only and therefore inert whenever a backend is selected, a control whose effect depends on the value of the control beneath it.

**The end state is one three-valued control**, `searchSource: none | native | <backendId>`:

```
Web search:  ( ) None
             ( ) The Provider's built-in search    ← offered only when providerHasNativeSearch
             ( ) SearXNG (searx.web)
```

That is **PR3.5**, promoted out of the optional follow-up list. It carries the migration (`search_source` + backfill, old columns kept one release), the `resolveSearchMode` rewrite, the form collapse, and the ADR Consequences edit. ~15 files.

Why it is not folded into this PR: that would put a data migration with a backfill, a change to what a searching turn resolves to, and a new frontend feature carrying a security check (the `javascript:` pill filter) into **one review** of ~22–25 files — and you split #384 at a much finer grain than that. The selector here is not throwaway: under PR3.5 it is still a `<Select>` fed by this route through the same three `formData` touch points, so most of it survives; what changes is the option list and the bound field. The Switch it replaces already exists on `main`.

**If you would rather have the collapse now and skip the interim shape, say so** — that is the main reason this is a draft.

## Found while testing: turn-level token growth (not this PR's bug)

Worth recording because the measurement only exists because this stack made searching reachable. On the test box, **one** searching turn on an Agent reached **229,677 input tokens across 15 steps** (7 `web_search`, 11 `read_url`) and died on the step ceiling without answering; another reached 67,688 in 9. Every `read_url` result — 5 000–10 000 chars — stays in the transcript and is re-sent on every subsequent step, so growth is quadratic in reads.

**Not web-search-specific and not a regression**: a Trigger run with `fetchUrl` ×2 and `updateWidgetData` ×9 and no search at all hit 187,066. Core's per-call caps already bound what one call contributes; what is unbounded is what is _retained_. The portable fix is clearing stale tool results from the transcript (deleting, not summarising — the two compose, and a cleared result is visibly gone where a summary is lossy in a way the model cannot detect). I would file it against the compaction workstream rather than this stack unless you want it here. Happy to open an issue with the numbers if useful.

## Open questions — answered 2026-08-08, kept for the record

All five are closed in the review thread. Left here rather than deleted, since the reasoning below is what those answers were given against: the read posture stays as written, the relabel stays deferred to PR3.5, PR3.5 goes after this, the `@platypus/schemas` lift is accepted, and the marker design is accepted without empirical OpenRouter verification.

1. **The catalog's read posture.** `requireAuth` + `requireOrgAccess()`, matching `GET /backends`, not `GET /plugins`' admin-only gate. Admin-only would break a real case: a workspace-scoped Provider is editable by a non-admin Workspace Owner when `providerSelfManagement` is set, and that owner would get an empty dropdown with no explanation. The cost is that any org member can learn which search plugins the deployment has installed, which `plugins` restricts to admins. The two existing precedents already disagree with each other, so converging them is above this PR — but the choice should be visible rather than inherited silently. **Tell me if you want admin-only and I will take the empty-dropdown case instead.**
2. **The `nativeSearchEnabled` relabel is deferred, not dropped.** #401's out-of-scope note assigns it here. It is deferred because the field disappears in PR3.5, and relabelling a control that is about to be deleted means writing user-facing copy twice and the docs twice on the same surface. The interim mitigation is one sentence at the point of selection plus the visible warning. **Say so if you would rather have the relabel now.**
3. **PR3.5 before or after this?** See above. Default is after.
4. **`isPresentableUrl` in `@platypus/schemas`** reverses your #384 answer, on the condition that answer named (a second consumer). Flagging rather than assuming.
5. **OpenRouter's native search is unverified empirically** — no account. The discriminator is written not to depend on it (the marker is core-owned and true on every provider), but if you have a way to check it, scenario 19 is "a native Provider must render with the generic renderer, not the compact card".

**One correction to the switch's own description did ship**, which my plan had ruled out along with the relabel. Its description read "Turn off for endpoints that don't implement it (e.g. vLLM, Ollama, LiteLLM)" — after PR2 that instructs an Operator to disable the backend on precisely the endpoints this Extension point exists for. That is shipped text walking a user into the trap, not the cosmetic relabel, so I treated the "nothing else" line as drawn one clause too wide and added the exception to the sentence. Still no relabel.

Two related findings are deliberately **not** fixed here, since they die with the collapse: the `providerType !== "Bedrock"` condition hiding the switch now hides a control that gates a Bedrock Provider's backend (not a live bug — the column is `notNull().default(true)`, so only a deliberate API write reaches that state), and the switch is still shown on Providers where it is inert.

What did get fixed is the new warning pointing at a control that is not on screen: where the switch is not rendered, the note beside the select names the Provider API as the only way into and out of that state. The condition is untouched; a warning this PR introduces should not create the dead end.

## Docs

Three of the four are mandatory because this PR makes shipped text false:

- **`extending/index.mdx`** said "Selection is not on the Provider form yet", and the warning callout under it was built on that premise. Both rewritten; the callout now carries only the coupling that survives.
- **`building-with-platypus/chat.mdx`** listed Bedrock and OpenAI-in-chat-mode as reasons the toggle is absent. Both are now conditional on there being no backend.
- **`building-with-platypus/triggers.mdx`** described the Trigger's Web Search switch as "Model native web search (if supported by provider)". A Trigger's `search` reaches the same `resolveSearchMode`, so on a Provider with a backend it searches through it.
- **`concepts/providers.mdx`** gains a new `## Web search` section — the page where Provider fields are documented, and there is no `building-with-platypus` Providers page for CLAUDE.md's table to point at. #392 stripped every search mention from it, so this is new rather than an amendment. It states the Bedrock case exactly as the form does: the switch is not shown because it is on by default there, and if it was switched off through the API the selection still will not run.

`self-hosting/providers-and-auth.mdx` is untouched: this PR adds no field limit or env var that page carries. The `webBackend` 200-char bound and the backend-author caps stay with PR4's `extending/web-search-backends.mdx`, where a plugin author reads them — a length limit beside a dropdown nobody types into is noise.

### Automated gates

`pnpm typecheck` and `pnpm lint` clean workspace-wide. `pnpm test` on the branch tip, merged up to `main` @ `41f69b5` with nothing behind: backend **1638**, schemas **122**, plugin-sdk 7, examples 3, frontend **181 of 184**.

Two failure classes are **local to this Windows box and reproduce identically on a clean `main`**, listed rather than rounded away:

- The three frontend failures are all `context-meter`, asserting `/42,000/` against a `toLocaleString()` with no explicit locale — this machine is `cs-CZ`, so it renders `42 000` with a narrow no-break space. Both files involved (`components/context-meter.tsx` and its test) came in with #457 and are byte-identical to `main` here; nothing in this PR's 20 files touches them.
- `apps/docs` fails `docs-contract.test.ts`'s internal-links case (24 pass / 1 fail, same on `main`) — a CRLF artifact, where the heading regex (`^#{1,6}\s+(.*)$`) keeps a trailing `\r`, so every anchor in the tree mis-slugs and the whole check goes empty rather than failing on anything this PR wrote. `git config core.autocrlf` is `true` here and the index is LF, so Linux CI is unaffected.

Both anchors this PR adds (`/concepts/providers#web-search`, `/extending#web-search-backends`) were checked by hand and resolve to real headings. Making either test environment-tolerant is a standalone `chore`, not this stack's.

### Real-world verification, against a backend written for this PR

The automated suite mocks the executor, so none of it proves the seam works against third-party code. To have something real to point a Provider at, I wrote a backend out of tree and made it public: **https://github.com/MesoX/platypus-searx-backend** — a plain `webBackends` contribution, `web_search` over self-hosted [SearXNG](https://docs.searxng.org/) and `read_url` over [obscura](https://github.com/h4ckf0r0day/obscura) via CDP. No search API key and no third-party account, so the whole Extension point is reproducible on any box. It is the first third-party consumer of the contract shipped in #384/#393; everything in the "four decisions" section above was found by _using_ that seam, not by reading it.

**Deployment under test:** a self-hosted box running the full stack off this branch, Provider `Local` = OpenAI + `apiMode: "chat"` (the vLLM shape — a Provider with no native search, i.e. exactly the case this Extension point exists for), backend id `searx.web`, both services on the compose network. Backend logs tailed on the observability line (`{ backend, plugin, tool, durationMs, outcome }`) throughout.

**Checked by hand, each with a signal that distinguishes it from every other case:**

| Path                       | What was exercised                                         | Signal it actually worked                                                                                                                                                                                         |
| -------------------------- | ---------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Search, results            | globe on, open-ended news query                            | compact card with the query and a count; `<Sources>` row of **≤10** pills where SearXNG returned **24–26** raw hits, so core's `MAX_SEARCH_RESULTS` slice is visibly running rather than the backend's raw output |
| Search, answer, no results | a query SearXNG answers from an engine answer box          | card reports **0 results**, the answer renders, and there is **no Sources row at all** — a zero-result search is not an error                                                                                     |
| Search, answer + results   | a query returning both                                     | pills **and** the answer box together, which separates "answer plumbing works" from "answer plumbing works alongside results"                                                                                     |
| Citation persistence       | full page reload on a finished turn                        | card and pills survive from the stored message, i.e. the marker and the result shape both round-trip through the DB                                                                                               |
| `read_url`, short page     | a page under the default slice                             | complete content, **no** continuation hint                                                                                                                                                                        |
| `read_url`, long page      | a long Wikipedia article at the default `max_length`       | 5 000 chars plus `[Content truncated. Pass start_index=5000 to continue reading.]` and `next_start_index: 5000` — byte-identical to `fetchUrl`, so a model that learned one tool pages the other correctly        |
| Egress guard               | a model-supplied loopback URL into `read_url`              | blocked, the browser is **never called**, one warn line carrying the policy reason                                                                                                                                |
| Stale selection            | plugin removed from `PLATYPUS_PLUGINS`, Provider unchanged | form shows `searx.web (not installed)` rather than a blank select; a searching turn injects no tools and warns with `providerId`                                                                                  |
| The documented trap        | Native web search switched off with a backend selected     | globe absent entirely — the row-4 behaviour PR3.5 removes, confirmed as described rather than assumed                                                                                                             |

**Not covered by the manual pass, stated so it is not read as more than it is.** This was a basic-scenarios run, not the full matrix: the failure paths (SearXNG down, obscura down, a slow page hitting `outcome: "timeout"` rather than `error`), the boot-refusal paths (a mis-keyed plugin config, an over-ceiling `timeoutMs`), the hostile-backend stub returning `javascript:`/`data:` result URLs, and the over-1M-char content cap were **not** run against the live deployment — all of them are covered by unit tests instead. Neither was the native-vendor comparison, which needs a second Provider on OpenRouter (no account; see open question 5). The `javascript:` pill filter in particular is proven by frontend tests, not by a live hostile backend.

### New tests

- Route: plugin annotation, `null` for an unowned id, empty registry, readable by a non-admin member, 401 unauthenticated, 403 for a non-member, no mutation verb.
- `isPresentableUrl` in schemas, including scheme casing.
- Provider form: absent with no backends installed; options carry the plugin name; `None`'s suffix follows capability; a stored-but-uninstalled id is named; the coupling warning appears only when the switch is off and the select stays enabled; `webBackend` round-trips through a save and sends `null`, not `""`; **the field stays mounted after a stored backend is cleared on a deployment with nothing installed** — the latch.
- Chat message: a pill per result with the backend's title; **no pill and no `href` for a `javascript:` or `data:` URL**; unpresentable results are not counted, and the card reports 0 when none were; URL fallback for a missing title; one pill for a page two searches both returned; merged with native `source-url` parts; no Sources row on an error; compact card carrying the query instead of the result JSON.
- Chat message, the native/plugin split: a provider-executed part keeps the generic renderer and lifts no sources, while its `source-url` citations still render; an **unflagged, unmarked** part — the OpenRouter shape — does the same, **including while it is still streaming its input**; a marked part shows the card mid-call, streaming or not; a result stored before the marker still lifts its pills; a denied call claims no search in flight; a page named by both rows renders one pill, titled by the backend.
- Chat message, native pills: titled by the title the vendor sent; the URL used when that title is empty; **no pill for a native citation whose URL is not presentable**.
- Backend: `composeWebBackend` stamps the marker on **both** Tools — pinned on the side that writes the seam the rendering reads.
- `humanizeToolType`: underscores are word boundaries, so `web_search`, `read_url` and namespaced MCP names stop rendering as "Web_search".

One rename not called out in its commit message, for completeness: the form's "no backend" sentinel went from `__none__` to `__platypus_no_web_backend__`, since a backend id is an arbitrary string and a contribution registering the old value would have read as None. Never persisted — it is mapped to `""` at the control's edge.


MesoX:
@willdady Hi, I would love to get some input here, before I mark this ready for review. Thanks!

willdady:
Reviewed. Answering the open questions first so you can take it out of draft — then four changes I'd like before merge, all frontend, all small.

## Your open questions

1. **Catalog read posture — keep `requireOrgAccess()` as written.** `GET /backends` is the closer precedent: both are process-wide registries served under a scoped path, feeding a picker. I confirmed the case you're protecting is real — `routes/provider.ts:39,103,169` gate create/update/delete behind `requireWorkspaceConfigAccess("providerSelfManagement")`, so a non-admin Workspace Owner genuinely can edit a Provider. Handing that person a silent empty dropdown is a worse outcome than disclosing a list of installed software names. No change.

2. **Relabel stays deferred.** Agreed, for your reason: PR3.5 deletes the field, and writing the copy and the docs twice on the same surface buys nothing. The corrected switch description, the warning at the point of selection, and the three docs pages carry it in the meantime. You drew the "nothing else" line in the right place — the old description actively instructed Operators into the trap, which is not cosmetic.

3. **PR3.5 after, as you proposed.** The state table is disclosed rather than hidden, and folding a backfill migration, a change to what a searching turn resolves to, and a new frontend security check into one review is the thing #384 was split to avoid.

4. **`isPresentableUrl` in `@platypus/schemas` — accepted, and the reversal is the right call.** The condition my #384 answer named has been met. Worth stating explicitly for the record: keeping it in `web-backends/` does not yield one copy, it yields two, since `apps/frontend` depends on `@platypus/backend` for types only and importing the predicate would be a new runtime dependency direction. Two copies of a security check, one of them deciding what becomes an `href`, is the outcome to avoid. `providerHasNativeSearch` is the precedent.

5. **OpenRouter — the marker design is accepted and I don't need it verified empirically.** Core-owned and true on every provider is the right property to build on. But one branch of the discriminator contradicts it; see change 1.

## Changes before merge

**1. `web-search-tool.tsx:69` — the `input-streaming` fallback is inverted.**

The comment at `:48` justifies it with "a marker is attached only when the tool call is parsed". That isn't what `ai@7.0.48` does. The marker is attached on the _first_ streaming chunk:

```js
// ai/dist/index.js:8625
case "tool-input-start": {
  const tool2 = getOwn(tools, chunk.toolName);
  controller.enqueue({ ...chunk, ..., ...tool2?.metadata != null ? { toolMetadata: tool2.metadata } : {} });
```

and `:7089-7112` writes it onto the `input-streaming` part. So the branch is exactly backwards: our own calls are already marked and never reach it, and the only parts that _do_ reach it are the unmarked native ones it exists to exclude. A native OpenRouter search renders as the compact card mid-stream, then swaps to the generic renderer at `input-available`.

Cosmetic in effect, but the comment states something false about the SDK and it's the one branch of the four with no test. Please make it `return false`, correct the comment, and add the case (`input-streaming`, no marker → generic renderer).

**2. `chat-message.tsx:176` — native Source pills discard the title.**

`SourceUrlUIPart` carries an optional `title` (`ai/index.d.ts:1899`) and Anthropic sends one, but both the existing line and this PR render `title={part.url}`. So a native Provider shows raw URLs as pill labels while a backend Provider shows real titles — a visible inconsistency in the row users actually read, in the surface this PR is about. `title={part.title ?? part.url}` plus a test. Pre-existing, but you're on the line and it's one word.

**3. `web-search-tool.tsx:166-169` — the card's count is unfiltered.**

`results.length` is the raw array; the Sources row counts only `isPresentableUrl` survivors. A backend returning three results with two `javascript:` URLs renders "3 results — listed above as sources" above one pill, and with three unsafe URLs it says the same thing above no Sources row at all. This is the rule you already committed to on the other side — "a 'Used 3 sources' row that opens onto one pill is its own bug" — applied once. Count what you will render.

**4. `provider-form.tsx:1210` — the field unmounts mid-interaction.**

On a deployment with nothing installed and a stored `gone.searx`, the form correctly offers the stale id so it can be cleared. Choosing None empties `formData.webBackend`, the visibility condition goes false, and the whole `Field` disappears before the save — no confirmation, no undo short of a reload. The save is correct; the control evaporating in the middle of the recovery path is not, and it's the same category as concealing a stored value. Latch it: once shown, keep it shown for the life of the form.

## Follow-ups, not blocking

**Unify the search card across native and plugin, as its own PR.** The end state I want is that a reader cannot tell from the presentation which search ran. Today the plugin path is the better one on both counts — plugin pills carry titles where native pills show URLs (change 2 closes that half), and the plugin card shows the query and a count where native gets a raw JSON dump.

What I checked, so the follow-up is scoped honestly rather than discovered mid-review:

- **Anthropic** (`@ai-sdk/anthropic@4.0.12`) — input is `{ query }`, the same key core uses; output is `Array<{ type: "web_search_result", url, title, ... }>`. Fully readable, uniform card works.
- **OpenAI** (`@ai-sdk/openai@4.0.11`) — input is `{}`, the query lives in the output as `action.query`/`action.queries`, and the tool output carries **no results at all**, only a record of the action taken. OpenAI can share the card shell but can never show a count. That's the vendor's limit, not a design choice, and the follow-up should say so rather than leave a reader wondering.
- **OpenRouter** — still unverified. Unrecognised shapes must keep falling back to the generic renderer.

Worth noting for that PR's framing: once the card is uniform, the discriminator stops deciding what a reader sees and only decides citation-lifting, so a misclassification costs a pill rather than a whole wrong-looking card. That is a meaningful reduction in what the marker has to get right.

**The token-growth measurement — yes, please file it**, with the numbers, against the compaction workstream rather than this stack. You're right that it isn't this PR's bug and isn't search-specific; the Trigger run at 187,066 with no search at all settles that. But the measurement only exists because this stack made searching reachable, and it shouldn't be lost in a PR thread.

## Notes

The `searx` backend being public is genuinely useful — being the first third-party consumer of the #384/#393 contract is what makes the four decisions above credible, and the "not covered by the manual pass" list is the right way to report a manual run. No objection to it living out of tree and unadopted.

I have not run the suites; I'm taking the green result from the description. The SDK and provider-package claims above I checked in `node_modules` rather than from memory.

Mark it ready once the four are in.

MesoX:
Thanks — all four are in, as `739807f` and `3f5e086`. Two of them shipped slightly wider than you asked, flagged below rather than left for you to find.

## The four changes

**1. `input-streaming` — fixed as `return false`, comment corrected, test added.** Your reading of the SDK is right and I checked it rather than taking it: `ai@7.0.48` copies a Tool's `metadata` onto the `tool-input-start` chunk and the UI stream writes it onto the part, so our own calls are already marked while streaming and check 2 has them. The only parts that reached step 4 were the unmarked native ones it existed to exclude. The new case is an unmarked streaming part keeping the generic renderer, plus its opposite — a marked streaming part showing the card — since the pair is what pins the branch.

Worth recording as a Consequence for PR4 alongside the marker itself: the discriminator now depends on the SDK carrying that metadata **from the first streaming chunk**, not only at parse time. I did not write a test for that half — it would assert a third party's stream shape, and the failure mode if a future SDK stopped doing it is cosmetic (generic renderer mid-stream, correcting itself at `input-available`). What is pinned is the side we own, that `composeWebBackend` stamps it.

**2. Native Source pills are titled — with one deviation: `||`, not `??`.** `??` keeps an empty-string title, which renders a pill with no label at all — worse than the raw URL the change exists to replace — and the plugin path already falls back this way. Same intent, one more falsy case. Tests for both the vendor-supplied title and the empty-title fallback.

**This one grew, and I would rather you saw it than found it: native `source-url` parts now go through `isPresentableUrl` as well.** That is not in your ask and it is a behaviour change on a shipped native path, so the reasoning: titling the pill takes the URL off the screen, and the raw-URL label was the only thing that made a `javascript:` or `data:` citation visible. With the label gone, such a citation renders as an ordinary-looking link with nothing to give it away. The consequence stated plainly rather than implied — an unpresentable native citation is now **dropped** from the row and from its count, not degraded to unlinked text. No live loss, since Anthropic and OpenAI emit `http(s)` only, but it is core's predicate deciding what a vendor's citation renders as. **Say the word and I will scope it back to plugin results only.**

**3. The card counts what it renders.** Through `webSearchSources`, the same helper the row is built from, so the rule is applied once rather than twice. It counts **this call's own** results deliberately: two searches in one turn that both surface a page total one more than the row, which deduplicates message-wide. The alternative is a card reporting a number with nothing to do with the search it heads. Tests for a partly-unpresentable result set and for one where nothing survives.

**4. The selector field is latched.** Once the reader has used the control it stays mounted for the life of the form. Set from the selector's own handler rather than derived, so the latch cannot hold the field open for a reason the reader never caused — an edit is the only thing that turns the visibility condition false. Test covers the recovery path you described: stored `gone.searx`, nothing installed, choosing None.

## The answers

Taken as given, with one addition worth keeping for the record: your point that leaving `isPresentableUrl` in `web-backends/` yields **two** copies rather than one, since `apps/frontend` depends on `@platypus/backend` for types only, is a stronger argument than the one I made and is the one I would use next time the question comes up.

## Your two follow-ups

**The uniform search card** — recorded as its own PR with your three vendor findings attached, including that OpenAI's tool output carries no results at all and so can never show a count. That is the kind of thing that gets rediscovered mid-implementation if it is not written down. Your framing that a uniform card shrinks a misclassification from a wrong-looking card to a missing pill is recorded with it, since it is also the argument for not leaving it indefinitely.

**The token-growth measurement** — agreed it should not be lost in this thread. I think its home is #261 (chat context compaction) rather than a fresh issue: that PR is the compaction workstream you pointed at, it is mine and still open, and the numbers are direct field evidence for the feature it proposes rather than a new report. One distinction I would keep visible when I put them there: what the measurement argues for is **pruning** stale tool results, which is not what #261 does — #261 summarises, pruning deletes, and they compose rather than substitute. So it lands there as evidence plus a named follow-up, not as a claim that #261 already covers it. Tell me if you would rather have a standalone issue and I will file one instead.

## Two notes

The verification numbers in the description are updated. Two failure classes are local to my Windows box and reproduce identically on a clean `main`, listed rather than rounded away: three `context-meter` tests assert `/42,000/` against a `toLocaleString()` with no explicit locale, and this machine is `cs-CZ`, so it renders `42 000`; and the `docs-contract` internal-links case fails on CRLF. Neither touches this PR's files. Backend 1638, schemas 122, frontend 181 of 184.

Also noted that you did not run the suites. Given that, the two deviations above are the parts I would most want a second pair of eyes on — they are the changes where I did something other than what you asked.

MesoX:
The token-growth measurement is filed: **https://github.com/willdady/platypus/pull/261#issuecomment-5292873099** — #261 (`feat(backend): chat context compaction`), which is the compaction workstream you pointed at rather than a fresh issue.

It carries all three runs (the 229,677-token turn that died on the step ceiling, the 67,688 one that did not, and the Trigger run at 187,066 with no search at all), framed as justification for that PR and explicitly **not** as a request to grow its scope — since what the numbers most directly argue for is pruning stale tool results, and #261 summarises. Those compose rather than substitute, so it lands there as evidence plus a named follow-up. Still happy to raise the pruning work as its own issue if you would rather have it queued than living in that thread.

willdady:
Reviewed and merging as-is. Four review findings plus a docs nit are tracked as follow-ups rather than blocking this PR:

- #467 — MCP tools merged unprefixed, so an MCP `web_search` hijacks the Web-search card (needs a direction call: namespace at the merge site vs. ownership check in the predicate)
- #468 — "this backend will not run" warning shows on Providers with no backend selected
- #469 — plugin Web search shows no elapsed time on its tool header
- #470 — `webBackendEdited` latch not reset when the form switches providers
- #471 — docs state the field visibility rule too narrowly

Issue #404, opened 2026-08-02
`maxOutputTokens` is never set and is not configurable per agent


willdady:
## Summary

Platypus never passes an output token ceiling to the AI SDK, and there is no way to configure one. Agents silently inherit the provider default, which for Anthropic via the AI SDK is commonly 4096 tokens. This causes long tool calls to be truncated mid-generation.

## Detail

The `modelArgs` object built in `apps/backend/src/runs/agent-runner.ts` and passed to both `streamText` and `generateText` includes `model`, `messages`, `system`, `tools`, `stopWhen`, `abortSignal`, `temperature`, `topP`, `topK`, `frequencyPenalty`, `presencePenalty` and `seed` — but no `maxOutputTokens`.

A repo-wide search for `maxOutputTokens`, `maxTokens`, `max_tokens` and `maxCompletionTokens` returns zero results.

There is also no per-agent field: `agentBaseSchema` in `packages/schemas/index.ts`, the `agent` table in `apps/backend/src/db/schema.ts`, and `apps/frontend/components/agent-form.tsx` all expose `maxSteps` (default 15, `DEFAULT_AGENT_MAX_STEPS` in `apps/backend/src/services/chat-execution.ts`) but no token ceiling. Sub-agents in `apps/backend/src/tools/sub-agent.ts` have the same gap.

`packages/schemas/index.ts` already acknowledges this on `modelConfigSchema`:

> "This object is also the intended home for future per-model metadata (e.g. max input/output tokens) — out of scope here."

## Why it matters

The provider's output budget covers reasoning tokens, assistant prose and tool-call JSON combined. An agent writing a long document through a tool — for example `upsertSkill`, whose `body` field permits up to 50,000 characters — cannot get near that limit, because generation is cut off well before it. The schema advertises a capacity the runtime cannot deliver.

## Observed impact

An agent attempting `upsertSkill` with a ~13,000 character body failed repeatedly. A ~4,100 character body succeeded. Nothing in the schema or the database was the constraint: the Zod cap is `body: z.string().min(48).max(50000)` in `apps/backend/src/tools/skill-management.ts`, and the column is unbounded Postgres `text` (see migration `apps/backend/drizzle/0002_messy_shotgun.sql`). The generation was truncated before the arguments were complete.

## Suggested fix

- Set an explicit, generous default `maxOutputTokens` in `modelArgs`.
[...]

PR #428, opened 2026-08-08
fix: mark a streamed reply that stopped at the model's output limit

### LINKED ISSUE #420: A streamed response truncated at the output limit still stops mid-sentence with nothing said
Split out of #406, which fixed this for unattended runs only.

## Summary

When a streaming Chat turn ends with a unified finish reason of `length`, the assistant's answer simply stops — mid-sentence, mid-list, mid-word. Nothing in the UI says it was cut off. The operator now sees it in the logs (#419 added the warning), but the person reading the reply does not.

The unattended path is already handled: `generate()` appends an explicit notice to the text it returns, so a Trigger or an Agent consuming that output can tell a truncated answer from a complete one. A user watching a stream gets nothing.

## Why #419 stopped here

The truncation notice needs a seam for putting text into a stream that has already been flushed to the client, and there isn't an obvious one:

- `onError` on the UI message stream is for errors. A truncated generation is not an error — the SDK resolves normally.
- `ChatMessageMetadata` (`apps/backend/src/types.ts`) has no frontend consumer at all, so adding a flag there today would be plumbing nothing reads.

Rather than invent a channel, #419 left this open deliberately.

## The seam that does exist

`toUIMessageStream`'s `messageMetadata` callback receives `{ part: TextStreamPart }` and **is called on the `finish` part**, which carries both `finishReason` and `rawFinishReason`. So the backend half is small: widen `ChatMessageMetadata` and populate it there, next to the `agentId` it already sets.

The real work is the other half — deciding how a truncated reply should *look*. Options worth weighing:

- A quiet inline marker at the end of the message ("response cut short at the model's output limit").
- A badge or callout on the message, consistent with how other non-fatal conditions are shown.
- Persisting the flag so it survives a reload, rather than only appearing live.

That is a product decision, not a mechanical one, which is why it wants its own issue rather than being bolted onto the instrumentation work.

## Acceptance

- [ ] A streaming turn that ends with `finishReason: 'length'` is visibly marked as truncated in the chat UI
- [ ] The marking survives a page reload (i.e. it is persisted, not stream-only)
- [ ] A turn that ends cleanly is unchanged — no marker, no extra metadata noise

## Not in scope

Adding a configurable output ceiling. #404 was closed as superseded and the reasoning there still holds: on Bedrock no ceiling is sent at all, and any global number below the model maximum would *lower* the ceiling Agents get today.

willdady:
Closes #420

## The problem

A streaming Chat turn ending with a unified finish reason of `length` stopped mid-sentence with nothing said to the reader. The operator saw it in the log (#419), and an unattended caller got `TRUNCATED_BY_TOKEN_LIMIT` appended to the returned text — but the person watching the stream got no signal at all.

## The change

`toUIMessageStream`'s `messageMetadata` callback runs on the terminal `finish` part, which carries the finish reason. That is the one channel that reaches a stream already flushed to the client, so the flag rides it.

- **`ChatMessageMetadata`** gains `truncatedByTokenLimit?: true`. `agentId` becomes optional: a direct provider/model Chat resolves no agent and previously yielded no metadata at all, so there was no way to flag a truncated direct chat. The avatar consumer already tolerated an absent `agentId`.
- **The metadata callback** now reads the stream part — `agentId` at `start`, the truncation flag at `finish`. Each event contributes only the key it owns, so the client-side merge leaves the attribution standing and a truncated Agent turn keeps its avatar.
- **It keys off the terminal `finish`, not `finish-step`.** A step inside a tool loop can end at the ceiling in a run that then completes normally; flagging those would mark answers that were never cut short. The per-step warn log is unchanged.
- **The chat renders** a muted line with a warning icon after the message, from an exported constant so tests assert the wording without restating the prose.

## Persistence

Verified rather than built — **no schema migration and no new column**. The `finish` chunk's metadata merges into the message before the stream's end callback fires, so `ChatSink` writes the flag into the existing `messages` `jsonb`, and the chat client re-seeds it from that row on load. This is the same path the Agent avatar has survived a reload on since #181.

## Tests

Backend covers the metadata callback at `start`, at a truncated `finish`, at a clean `finish`, on a mid-loop `finish-step`, and for a direct provider/model stream with no attribution. Frontend covers the marker rendering, not rendering, and the Agent avatar surviving a truncated turn.

`pnpm lint`, `pnpm typecheck` and `pnpm test` pass.

## Docs

A **When a reply stops early** section on the Chat page, shipped in this PR: what the marker says, that it survives a reload, and a warning that **Regenerate** re-runs the whole turn rather than carrying on from where the reply stopped.

## Not in scope

A configurable output ceiling (#404 was closed as superseded and the reasoning holds), any retry or "continue the response" affordance, changes to the unattended notice, and surfacing any finish reason other than `length`.

## Note for review

The brief asked for the marker "after the last text part". It renders after the whole message instead: if a turn is cut off mid tool-call, the marker belongs at the end rather than above the tool block. In the common case — text last — the two are identical.

### DIFF
--- a/apps/backend/src/runs/agent-runner.test.ts
+++ b/apps/backend/src/runs/agent-runner.test.ts
@@ -52,6 +52,11 @@ const {
       onStepFinish: undefined as ((step: unknown) => void) | undefined,
       onFinish: undefined as
         ((ctx: { messages: unknown[] }) => Promise<void> | void) | undefined,
+      // `toUIMessageStream`'s metadata extractor. The SDK calls it per stream
+      // part; the tests call it by hand with the part they care about.
+      messageMetadata: undefined as
+        | ((opts: { part: { type: string; finishReason?: string } }) => unknown)
+        | undefined,
       responseSentinel: { __isResponse: true },
     },
   };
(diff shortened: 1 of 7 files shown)


willdady:
Superseded by #429 — same commit, branch renamed to `fix/420-mark-truncated-streamed-reply` to match the change type.

PR #451, opened 2026-08-11
fix(backend): validate tool set contributions

### LINKED ISSUE #444: Tool set Contributions get no validation; a missing id registers as `<plugin>.undefined`
## Summary

The plugin loader validates Sandbox backend and Web-search backend Contributions before registering them. Tool set Contributions get no validation at all — the loop's first statement destructures the untrusted value and passes its `id` straight into the namespacing helper.

`apps/backend/src/plugins/loader.ts:381-383`:

```ts
for (const contribution of manifest.contributes.toolSets ?? []) {
  const { id, name: tsName, category, description, tools } = contribution;
  const effectiveId = contributionId(id);
```

Per-Contribution checks by Extension point:

| Loop | Lines | Checks |
| --- | --- | --- |
| `sandboxBackends` | `:425-548` | 6 — non-object, empty `backend`, empty `name`, missing `create`, `configSchema` factory throw, duck-typed `safeParse` on both schemas |
| `webBackends` | `:552-653` | 5 — non-object, empty `backend`, empty `name`, missing `createExecutors`, `timeoutMs` range |
| `toolSets` | `:381-421` | **0** |

Consequences for a third-party plugin with a malformed Tool set Contribution:

- a missing `id` mints `"<plugin-name>.undefined"` via `contributionId` (`:376-377`) and registers it — it then appears in the catalog and in `GET /plugins` as a real Tool set;
- a `null` entry throws an unattributed `Cannot read properties of null` at the destructure, with no indication of which plugin caused it;
- a missing `tools` registers a Tool set that resolves to nothing at Chat-turn time.

## The rationale is already written down — against the other two loops

Both hardened loops carry a comment describing precisely this failure, and neither was applied back to the Tool set loop:

`:428-432` (sandbox)
> Each malformed shape used to surface far from its cause: `null` as an unattributed `Cannot read properties of null`, a missing `backend` as a plausible-looking `"acme.undefined"` in the catalog…

`:563-566` (web)
> Identity is checked before it is namespaced: `contributionId(undefined)` would otherwise mint a plausible-looking `"acme.undefined"` and register it…

Tool sets are the most-contributed Extension point — 11 across `@platypus/tools-basic`, `@platypus/tools-platform` and `@platypus/web-fetch`, plus one from each of the two example third-party plugins — and the only one with no contribution validation.

## Suggested direction

Give the Tool set loop the same fail-loud, plugin-attributed checks the other two have: non-object entry, non-empty `id`, non-empty `name`, and a usable `tools` (object or function), all validated **before** `contributionId` runs.

The three loops now hold three near-identical implementations of the same job, which is how the drift keeps happening — worth considering whether validation and registration belong in one place rather than one copy per Extension point.

_Surfaced by an architecture review of the Plugin / Extension point machinery._

floze-the-genius:
## Summary

- validate every Tool set Contribution before id namespacing and registration
- abort startup with a plugin-attributed error for malformed entries, blank identity fields, or invalid tools
- document the runtime validation and fail-loud self-hosting contract

Fixes #444

## Validation

- RED: the new loader regression table failed before the production change; malformed entries either threw an unattributed destructuring error or reached registration
- `pnpm --filter @platypus/backend test loader` — 76 passed
- `pnpm typecheck` — 7/7 tasks passed
- `pnpm test` — 8/8 tasks passed; backend 1574 passed
- `pnpm --filter @platypus/docs test` — 24 passed
- `pnpm build` — 5/5 tasks passed, including all 46 documentation pages
- targeted Prettier and `git diff --check` passed

Local validation used Node 26.5.0 and pnpm 10.26.1; repository CI remains the pinned Node 24 check.

### DIFF
--- a/apps/backend/src/plugins/loader.test.ts
+++ b/apps/backend/src/plugins/loader.test.ts
@@ -357,6 +357,66 @@ describe("loadPlugins", () => {
     ).rejects.toThrow(/@empty\/plugin.*manifest/s);
   });
 
+  it.each([
+    [
+      "a non-object entry",
+      null,
+      /@platypus\/bad.*tool set contribution must be an object/s,
+    ],
+    [
+      "a string entry",
+      "broken",
+      /@platypus\/bad.*tool set contribution must be an object/s,
+    ],
+    [
+      "an array entry",
+      [],
+      /@platypus\/bad.*tool set contribution must be an object/s,
+    ],
+    [
+      "a missing id",
+      { name: "Broken", category: "Test", tools: {} },
+      /@platypus\/bad.*non-empty "id"/s,
+    ],
+    [
+      "a missing name",
+      { id: "broken", category: "Test", tools: {} },
+      /@platypus\/bad.*tool set "broken".*non-empty "name"/s,
+    ],
+    [
+      "missing tools",
+      { id: "broken", name: "Broken", category: "Test" },
+      /@platypus\/bad.*tool set "broken".*"tools" object or function/s,
+    ],
+  ])(
+    "aborts (fail-loud, plugin-named) on %s",
+    async (_label, contribution, expected) => {
+      const { register, calls } = makeRegister();
+      await expect(
+        loadPlugins({
+          pluginNames: ["@platypus/bad"],
+          builtinPlugins: {
+            "@platypus/bad": () =>
+              Promise.resolve({
+                plugin: {
+                  name: "@platypus/bad",
+                  version: "0.1.0",
+                  apiVersion: 1,
+                  contributes: {
+                    toolSets: [
+                      contribution,
+                    ] as unknown as ToolSetContribution[],
+                  },
+                },
+              }),
+          },
+          register,
+        }),
+      ).rejects.toThrow(expected);
+      expect(calls).toHaveLength(0);
+    },
+  );
+
   it("aborts (fail-loud) on a duplicate id, naming both owning plugins", async () => {
     const { register } = makeRegister();
     // Two core plugins keep bare ids, so a shared `time` id collides directly.
--- a/apps/backend/src/plugins/loader.test.ts
+++ b/apps/backend/src/plugins/loader.test.ts
@@ -379,13 +379,33 @@ describe("loadPlugins", () => {
       /@platypus\/bad.*non-empty "id"/s,
     ],
     [
+      "a whitespace-only id",
+      { id: "   ", name: "Broken", category: "Test", tools: {} },
+      /@platypus\/bad.*non-empty "id"/s,
+    ],
+    [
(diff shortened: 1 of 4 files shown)


willdady (commented):
Nice catch on the gap: `toolSets` was the only one of the three contribution loops that trusted its entries, and the fix lands in the right place — before `contributionId()` and before `register()`, so a rejected contribution leaves nothing in the owner map or the registry. The `it.each` table pinning `calls.length === 0` is the right assertion. Rejecting arrays is also an improvement over the sandbox/web loops, which let an array through their `typeof === "object"` check.

One thing to fix before merge:

**`category` is required but unvalidated** (`apps/backend/src/plugins/loader.ts:411`). `ToolSetContribution` declares `category: string` as required (`packages/plugin-sdk/index.ts:110`), but the new checks cover only `id`, `name`, and `tools`. A JS plugin that omits it boots fine, registers, and the value flows through `agent-discovery.ts:62` into an API response whose `ToolSet.category` is typed `string` (`apps/backend/src/tools/index.ts:17`) — the frontend then groups that tool set under `undefined`. That's the same "plausible-looking value in the catalog" failure the new comment cites as the reason for the other checks. Worth noting the sandbox and web loops leave no required field unchecked, so this is the one spot where the new code diverges from the pattern it's mirroring.

```ts
if (
  typeof contribution.category !== "string" ||
  contribution.category.trim() === ""
) {
  throw new Error(
    `Plugin "${manifest.name}": tool set "${contribution.id}" must declare a non-empty "category".`,
  );
}
```

Related: `extending/index.mdx` says "Every Contribution is validated before it is registered" and then lists only `id`/`name`/`tools`. Adding the `category` check makes that sentence true, which seems like the better resolution than softening the docs.

Optional, take or leave:

- **`id` is trim-tested but used untrimmed.** `" my set "` passes `.trim() !== ""` and is then used verbatim, producing `acme. my set ` — which gets persisted into `agent.toolSetIds`. The other half of that composed id, the manifest name, is held to a url-safe slug at `loader.ts:357` precisely because it becomes part of the id. Pre-existing for the backend loops, so a follow-up is fine if you'd rather do it uniformly.
- **Errors don't say which entry failed.** On a plugin contributing five tool sets, "every tool set contribution must be an object" gives an Operator nothing to grep for. `.entries()` plus an `at index N` suffix would help — though it does match how the existing loops behave.
- **Two test cases missing from the table:** a whitespace-only `id` (`"   "`), which is the branch `.trim()` exists for and is currently unpinned, and `tools: []`, given `Array.isArray` was added deliberately for it.

Docs shipping in the same PR is appreciated, and the `self-hosting/configuration.mdx` rewrite reads at the right altitude for an Operator.

willdady (approved):

PR #455, opened 2026-08-12
fix(backend): propagate tool set factory errors

### LINKED ISSUE #443: A Tool set factory that throws is silently reinterpreted as "not a Tool set"
## Summary

`loadTools` wraps both the Tool set lookup **and** the invocation of the tool set's factory in one bare `catch` with no error binding. Any failure inside a Tool set factory is therefore swallowed and reinterpreted as "this id isn't a static Tool set, go look for an MCP with the same id" — after which the turn logs that the id is neither, and proceeds with no tools from it.

`apps/backend/src/services/chat-execution.ts:996-1011`:

```ts
for (const toolSetId of agent.toolSetIds) {
  try {
    const toolSet = getToolSet(toolSetId);
    const resolvedTools =
      typeof toolSet.tools === "function"
        ? await toolSet.tools({ workspaceId, agentId: agent.id, orgId, frontendUrl, userId: userId || "" })
        : toolSet.tools;
    Object.assign(tools, resolvedTools);
  } catch {
    // Static tool set not found — fall back to MCP lookup.
    const mcp = await queries.getMcp(toolSetId, orgId, workspaceId);
    ...
```

The comment states the intended scope of the `catch` — lookup failure — but the `try` covers three more lines, including an `await` on plugin-contributed code.

Observed symptom: an Agent silently loses a Tool set's tools, and the only trace is
`Tool set with id '<id>' not found as static tool set or MCP`, which points at a
registration problem that isn't the real cause. The original error is discarded entirely.

## Why this one registry and not the others

The three Extension point registries disagree on how to signal a lookup miss:

| Registry | Miss behaviour |
| --- | --- |
| `apps/backend/src/tools/index.ts:51-56` | **throws** |
| `apps/backend/src/sandbox/index.ts:46-48` | returns `undefined` |
| `apps/backend/src/web-backends/index.ts:83-85` | returns `undefined` |

Because `getToolSet` signals by throwing, its caller needs a `try`/`catch` for control flow, and the factory call ended up inside it. The two registries that return `undefined` let their callers branch on the value (`tools/index.ts:97-103`, `chat-execution.ts:472-492`) and structurally cannot have this bug.

## Suggested direction

Narrow the `try` to the lookup alone so a factory throw propagates (or is logged with its error), e.g. by having the miss return `undefined` and branching on it — which would also bring the Tool set registry in line with the other two.

Part of a broader drift between the three registries; a companion issue covers the missing validation on Tool set Contributions.

_Surfaced by an architecture review of the Plugin / Extension point machinery._

kudala-bharani:
## Summary

- limit the static tool-set lookup catch to registry misses
- preserve MCP fallback for unknown tool-set IDs
- propagate tool-set factory failures with their original error
- add regression coverage proving factory failures do not query MCPs

## Validation

- `pnpm --filter @platypus/backend test chat-execution.test.ts`
- `pnpm test`
- `pnpm typecheck`
- `pnpm lint`
- Prettier check on changed files

Fixes #443

### DIFF
--- a/apps/backend/src/services/chat-execution.test.ts
+++ b/apps/backend/src/services/chat-execution.test.ts
@@ -98,6 +98,7 @@ import {
   composeWebBackend,
   registerWebBackend,
 } from "../web-backends/index.ts";
+import { registerToolSet } from "../tools/index.ts";
 import { logger } from "../logger.ts";
 import { FileValidationError } from "./file-gate.ts";
 import { resetExtractedTextCache } from "./file-extraction.ts";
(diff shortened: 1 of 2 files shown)


willdady (approved):
@kudala-bharani Thank you!

PR #459, opened 2026-08-13
fix(backend): resolve sub-agent maxSteps to the shared default

### LINKED ISSUE #441: A Sub-Agent runs 50 steps where the same Agent runs 15
## Summary

`agent.maxSteps` is a plain nullable integer with no database default (`apps/backend/src/db/schema.ts:235`). The two paths that read it disagree about what `null` means:

- **Chat turn** — `apps/backend/src/services/chat-execution.ts:938`
  ```ts
  resolvedMaxSteps = agent.maxSteps ?? DEFAULT_AGENT_MAX_STEPS;   // 15, defined at :64
  ```
- **Sub-Agent delegation** — `apps/backend/src/tools/sub-agent.ts:400`
  ```ts
  maxSteps: subAgent.maxSteps || 50,
  ```

So one Agent row gets a 15-step ceiling when a user selects it on a Chat, and a 50-step ceiling when a parent Agent delegates to it — 3.3× the budget, silently, from the same configuration.

Two smaller problems ride along:

- `||` rather than `??` means an explicitly configured `maxSteps: 0` also becomes 50.
- `DEFAULT_AGENT_MAX_STEPS` is documented at `chat-execution.ts:58-63` as mirroring the new-agent create-form default. The delegate path's `50` is a bare literal appearing twice (`sub-agent.ts:126` and `:400`) with no stated rationale.

## Why it happens

The Sub-Agent path does not go through `prepareChatTurn` or `AgentRunner` — it builds its own `ToolLoopAgent` (`sub-agent.ts:147`) and resolves the Agent row's fields itself. Anything the two paths both need has to be written twice and kept in step by hand.

`agent-runner.ts:178` already names the fix as future work:

> Out of scope (later PRs): Sub-agent runs as AgentRunner consumers (PR #4)

## Suggested direction

Short term, make the two paths agree — export the default from one place and use `??` on both. Which value is correct is a product call: 15 is the form default a user would expect to apply, but a delegate doing multi-step research plausibly wants more headroom. If the answer is "more headroom", that should be an explicit, documented Agent-level default rather than a literal that only applies when a parent invoked it.

Longer term this is one symptom of the Sub-Agent path being a parallel implementation of a Chat turn; see the sibling issues on truncation and the run lifecycle.

_Surfaced by an architecture review of the Chat turn execution path._

lesbass:
## Summary

`agent.maxSteps` is a nullable column with no DB default. The Chat-turn path resolves an unset value to `DEFAULT_AGENT_MAX_STEPS` (15), but the Sub-Agent delegation path used a bare `50` literal and `||` instead of `??`. The same Agent row therefore got a 15-step ceiling on a Chat and a 50-step ceiling when delegated to — and an explicit `maxSteps: 0` also silently became 50.

## Changes

- Import the existing `DEFAULT_AGENT_MAX_STEPS` export (`apps/backend/src/services/chat-execution.ts`) into `apps/backend/src/tools/sub-agent.ts`.
- Replace the duplicated `50` literal in both the `createSubAgentTool` default and the `createSubAgentTools` resolution with `DEFAULT_AGENT_MAX_STEPS`.
- Use `??` instead of `||` so an explicit `maxSteps: 0` is honored as a zero-step ceiling.

No change to the numeric default value (15) or to the create-form default/validation.

## Verification

- `pnpm --filter backend typecheck` — passes
- `pnpm --filter backend lint` — passes
- `pnpm --filter backend test` — 112 files, 1611 tests pass (includes new regression tests asserting `maxSteps: null` → default and `maxSteps: 0` → 0)

Closes #441

### DIFF
--- a/apps/backend/src/services/chat-execution.ts
+++ b/apps/backend/src/services/chat-execution.ts
@@ -31,6 +31,7 @@ import {
 } from "./memory-retrieval.ts";
 import {
   aliasNameFromReference,
+  DEFAULT_AGENT_MAX_STEPS,
   providerHasNativeSearch,
 } from "@platypus/schemas";
 import type { ConcreteModelId, Provider, Skill } from "@platypus/schemas";
@@ -55,14 +56,6 @@ import {
 } from "./file-gate.ts";
 import type { PlatypusUIMessage } from "../types.ts";
 import { listScopedByIds, resolveScoped } from "./scoped-resource.ts";
-
-/**
- * Default agentic step ceiling for an agent that has no explicit `maxSteps`.
- * Mirrors the new-agent create-form default. Keeps API-created agents sane
- * (a single step never lets a tool-calling agent finish its work) while
- * staying low enough to bound a model that fails to converge.
- */
-export const DEFAULT_AGENT_MAX_STEPS = 15;
 
 // --- Errors ---
 
--- a/apps/backend/src/tools/agent-management.ts
+++ b/apps/backend/src/tools/agent-management.ts
@@ -34,7 +34,12 @@ export function createAgentManagementTools(
       providerId: z.string().describe("Provider ID to use"),
       modelId: z.string().describe(MODEL_ID_DESCRIPTION),
       instructions: z.string().optional().describe(INSTRUCTIONS_DESCRIPTION),
-      maxSteps: z.number().optional().describe("Max agentic steps"),
+      maxSteps: z
+        .number()
+        .int()
+        .min(1)
+        .optional()
+        .describe("Max agentic steps, at least 1"),
       temperature: z.number().optional().describe("Sampling temperature"),
       topP: z.number().optional().describe("Top-p sampling"),
       topK: z.number().optional().describe("Top-k sampling"),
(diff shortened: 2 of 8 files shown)


willdady:
Thanks for this — the diagnosis was right and the `50` literal shouldn't have been there. I've pushed four commits onto your branch rather than sending you round again.

The one thing I'd push back on is the `??`. It does make `maxSteps: 0` reach the SDK, but `stepCountIs(n)` compiles to `({ steps }) => steps.length === n`, and the loop only evaluates stop conditions after a step has completed — so `steps.length` is never below 1 when it's checked. `stepCountIs(0)` therefore never fires, and a "zero-step" Agent ends up with no ceiling at all, looping as long as the model keeps calling tools. Unlike the Chat path, sub-agents have no `noProgress` backstop to catch it. So the test asserting a zero-step ceiling was pinning something the SDK doesn't actually implement, and I've dropped it.

The real gap was upstream: `agentSchema.maxSteps` had no bound, so `0` (and negatives, and fractions the integer column can't hold) could be written through both the REST route and the agent-management tool. That's now `int().min(1)`, matching the `min="1"` the Agent form already had, so the value fails at the edge instead of silently removing the ceiling at run time. `??` stays — it's the right operator once no falsy-but-valid value exists.

The other three commits are follow-on tidying your change made possible: your import created a `sub-agent.ts` ↔ `chat-execution.ts` cycle (harmless today, but it drags the whole DB/provider graph into a leaf tool module and would break the moment either side used the constant at module scope), so `DEFAULT_AGENT_MAX_STEPS` now lives in `@platypus/schemas` next to the bound it pairs with, and the frontend's two hardcoded `15`s read from it too.

Docs updated in the same PR per our contributing rules. Green on tests, typecheck, and lint. Thanks again for spotting it.

PR #515, opened 2026-08-15
fix: namespace MCP tools to prevent built-in collisions

### LINKED ISSUE #467: fix: MCP tools are merged unprefixed, so an MCP `web_search` hijacks the Web-search card
Found while reviewing #412. Not introduced by that PR's core mechanics, but #412 is what makes it reachable.

## The gap

MCP tools are merged into the tool map under their raw server-side names, with no namespacing:

```ts
// apps/backend/src/services/chat-execution.ts:1015
const mcpTools = await mcpClient.tools()
Object.assign(tools, mcpTools)
```

Nothing stops an MCP server from exposing a tool named `web_search` (or shadowing any other built-in name).

On the frontend, `isPluginWebSearchPart` (`apps/frontend/components/web-search-tool.tsx:66`) has a fallback branch that matches on shape alone — `output-available` plus a `results` array or an `error` string — with no ownership check.

## Impact

An MCP server exposing a tool named `web_search` that returns `{ results: [...] }`:

- renders on the compact Web-search card, which never renders tool output — so the tool's real payload silently disappears from the transcript
- has its result URLs lifted into the message's Sources row and presented as citations

## Two places this could be fixed

1. **Namespace MCP tools at the merge site.** Closes the whole class of collision, not just this one. Cost: changes the tool names the model sees and the names written into stored message parts, so it needs a decision on the prefix format and on what happens to existing chat history.
2. **Give `isPluginWebSearchPart` an ownership check** so only parts carrying `WEB_BACKEND_TOOL_MARKER` (or a known-native tool name) take the card. Narrow and cheap, but leaves the underlying shadowing gap open.

Filed for a direction call — (1) is the real fix but is the wider change.

tagadearpit:
## Summary

Fixes #467.

MCP tools are now namespaced before they enter the chat tool map as `mcp__<tool-set-id>__<tool-name>`. This prevents an MCP server from shadowing a built-in tool such as `web_search`, so a server-provided payload cannot be misclassified by the Web-search card or have its URLs presented as native citations.

The change is isolated to MCP session resolution. Built-in tool precedence remains unchanged, and MCP tools continue to be normalized and disposed through the existing session lifecycle. The new names are model-visible and will be used in newly stored tool parts; existing stored history is not rewritten.

## Reproduction

1. Configure an MCP server that exposes a tool named `web_search` and returns a payload shaped like `{ "results": [...] }`.
2. Assign that MCP server to an agent and start a chat turn.
3. Before this change, the raw MCP tool name is merged as `web_search`, which collides with the native Web-search tool.
4. The frontend shape fallback can then render the MCP result as the compact native Web-search card and lift its URLs into the message Sources row, while the MCP payload itself is not rendered as a generic tool result.
5. After this change, the tool is exposed as `mcp__<tool-set-id>__web_search`; it cannot overwrite the native `web_search` entry and is rendered as an ordinary MCP tool result.

## Changes

- Namespace every MCP tool with its stable tool-set ID before collision reporting and insertion into the turn map.
- Update tool-session and chat-turn regression coverage, including nested sessions and organization-scoped MCP resolution.

## Verification

- `pnpm format`
- `pnpm --filter @platypus/backend lint`
- `pnpm --filter @platypus/backend typecheck`
- `pnpm --filter @platypus/backend exec vitest run --maxWorkers=1`

The complete backend suite passed: 120 test files and 1,818 tests. The repository root lint attempt was terminated by the sandbox under memory pressure; the affected backend lint gate passed serially. The checkout uses Node 22 while the repository declares Node >=24, and pnpm reports that engine warning, but the targeted and full backend checks completed successfully.

### DIFF
--- a/apps/backend/src/services/chat-execution.test.ts
+++ b/apps/backend/src/services/chat-execution.test.ts
@@ -1269,7 +1269,7 @@ describe("chat-execution", () => {
         queries,
       );
 
-      expect(turn.stream.tools).toHaveProperty("mcpTool");
+      expect(turn.stream.tools).toHaveProperty("mcp__mcp-1__mcpTool");
       await turn.dispose();
     });
 
@@ -1294,7 +1294,7 @@ describe("chat-execution", () => {
         queries,
       );
 
-      expect(turn.stream.tools).not.toHaveProperty("mcpTool");
+      expect(turn.stream.tools).not.toHaveProperty("mcp__mcp-1__mcpTool");
       await turn.dispose();
     });
 
@@ -1324,7 +1324,7 @@ describe("chat-execution", () => {
         queries,
       );
 
-      expect(turn.stream.tools).not.toHaveProperty("mcpTool");
+      expect(turn.stream.tools).not.toHaveProperty("mcp__mcp-1__mcpTool");
       await turn.dispose();
     });
   });
--- a/apps/backend/src/tools/tool-session.test.ts
+++ b/apps/backend/src/tools/tool-session.test.ts
@@ -187,7 +187,7 @@ describe("openToolSession", () => {
       return close;
     };
 
-    it("serves an MCP server's tools for an id no Tool set claims, and closes it on dispose", async () => {
+    it("serves namespaced MCP tools for an id no Tool set claims, and closes it on dispose", async () => {
       const close = connected({ mcpTool: toolNamed("mcpTool") });
 
       const session = await openToolSession(
@@ -196,7 +196,7 @@ describe("openToolSession", () => {
         queriesFor([mcpRow()]),
       );
 
-      expect(session.tools).toHaveProperty("mcpTool");
+      expect(session.tools).toHaveProperty("mcp__mcp-1__mcpTool");
       expect(close).not.toHaveBeenCalled();
 
       await session.dispose();
@@ -221,7 +221,7 @@ describe("openToolSession", () => {
         queriesFor([mcpRow()]),
       );
       const execute = (
-        session.tools.listItems as unknown as {
+        session.tools["mcp__mcp-1__listItems"] as unknown as {
           execute: (a: unknown, o: unknown) => Promise<unknown>;
         }
       ).execute;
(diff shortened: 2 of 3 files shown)


willdady:
Thank you for this.

First, the important part: your diagnosis is correct. An MCP server really can
expose a tool called `web_search`, it really does land in the tool map
unprefixed, and the result really does get picked up by the wrong renderer. The
patch is clean, the tests are sensible, and you found a genuine bug in a corner
of the code that isn't obvious.

Where we landed is that we're going to close this PR — not because the work is
wrong, but because #467 wasn't ready to be picked up, and that's our fault
rather than yours.

That issue is labelled `needs-triage` and ends with "Filed for a direction call",
laying out two possible fixes. Choosing between them was a maintainer decision we
hadn't made yet, and nothing in CONTRIBUTING.md told you that. We've opened #531
to fix that gap so the next person doesn't hit it.

## Why the namespacing route needs more than a patch

Working through your PR is actually what convinced us the decision needs writing
down first. A few things it runs into:

**It doesn't fix the reported symptom for existing chats.** Renaming tools helps
new conversations, but every chat already saved still renders the bad card,
because nothing on the display side changed. The original complaint stays
reproducible.

**Tool names get referenced by name, in places we don't control.** MCP servers
routinely write descriptions like "call `list_items` before this tool". More
importantly, our users write things like "always use `search_docs` first" in
their agent's Instructions — text saved in their database that we can't rewrite.
After a rename, both quietly point at tools that no longer exist.

**It can break setups that work today.** Providers cap tool names at 64
characters. `mcp__` plus a 21-character id plus `__` spends 28 of those, so any
MCP tool with a name over 36 characters would start failing. The tests use a
short fake id, so this doesn't show up in CI.

**A couple of surfaces would need to follow.** The chat currently renders the raw
name, and the MCP "test connection" screen still lists the unprefixed names, so
the two would disagree.

None of these are mistakes in your patch — they're the cost of the direction, and
they're exactly what triage should have surfaced before you spent an evening on
it.

Thanks again for digging into this properly rather than papering over it.

PR #516, opened 2026-08-15
feat(backend,frontend): collapse native search and Web-search backend into one searchSource control

MesoX:
# PR3.5 — Collapse the two search knobs into one `searchSource` control

Fourth in the #329 stack, behind #384 (PR1), #393 (PR2) and #412 (PR3). Branch `feature/provider-search-source-collapse`, merged up to `main` @ `f9ed50e` (2.7.1). Only PR4 — the ADR status flip — comes after it.

PR3 shipped the Web-search backend selector next to the `nativeSearchEnabled` switch, and the two fight over one slot: the switch's name predates Web-search backends but it gates them too, so switching it off to disable a native tool that never worked also silently kills a backend selected later. PR3 could only describe that trap, in a `text-destructive` note beside the select and a second one worded differently where the switch is not rendered. This removes it: **one three-valued field, one precedence list, no state where two stored values contradict each other.**

**28 files, +4860/−638** — but 4342 of those insertions are drizzle's `0051` snapshot. The hand-written change is **25 files, +518/−638**: a net deletion, because most of what goes was written to explain the coupling.

## The end state

```
Web search:  ( ) None
             ( ) The provider's built-in search   ← only when providerHasNativeSearch
             ( ) SearXNG (acme-search)
```

One click moves between native and a backend, and the capability question answers itself by whether the built-in option is in the list. On a vLLM Provider the list is None plus the contributed backends, which needs no explanation — so PR3's helper text, its coupling warning and its two-branch wording all go rather than being ported.

`resolveSearchMode` reads straight down the same list: `"none"` (or an absent value) serves nothing; anything that is not a reserved literal is a backend discriminator and serves that backend, ahead of native; `"native"` serves the provider's own tool when `providerHasNativeSearch`. `canSearch` in `chat.tsx` mirrors it exactly.

## The two decisions the plan said to settle before writing code

**1. The old columns stay for one release.** `0051` adds `search_source`, backfills from both old columns, and drops neither — a backfill that reads a column it removes in the same migration has no rollback path. They are unread by the app from this PR on; a later `chore` drops them.

That also buys the API a back-compat window. A caller still `POST`/`PUT`ing `nativeSearchEnabled` / `webBackend` is mapped onto `searchSource` by a `z.preprocess` on the create and update schemas, using the exact precedence the old code encoded inline (`nativeSearchEnabled: false` beat a set `webBackend`), so a legacy payload resolves to the same search behaviour it had before. The mapping applies only when the payload omits `searchSource` outright — a caller sending both is assumed to know what it's doing.

**Worth flagging, since it is the one place the window leaks:** the deprecated columns are still returned on read (the Provider read model is `$inferSelect`, the whole row). A legacy client that reads `nativeSearchEnabled` back therefore sees its pre-collapse value frozen, not a projection of `searchSource`. Writing a projection would mean re-deriving a two-field answer from a one-field source, which is the ambiguity this PR exists to delete; freezing them is the honest option for one release, and the `chore` that drops the columns ends it.

**2. `"native"` is backfilled uniformly, and the form coerces it on load.** Today `nativeSearchEnabled: true` + no `webBackend` on Bedrock means "no search" via the capability check, not via the stored value. Encoding the capability table in backfill SQL would put `providerHasNativeSearch` in a second language, so the migration backfills `"native"` for every such row and the Provider form coerces an unavailable `"native"` to `None` **on load, without auto-saving**. Resolution was already safe — the capability gate still applies — so the coercion is purely so the form does not render "the provider's built-in search" as selected on a Provider that does not offer it.

## What PR3 shipped that this deletes, and what survives

Deleted, all of it existing only because the two-field shape did:

- **The coupling warning** and the coupling clause in the select's description, including the two-branch wording that said a different thing where the switch is not rendered (#507's fix, and #472's gate on it).
- **The visibility rule and its `webBackendEdited` latch.** `searchSource` is always rendered — `None`, and where the Provider is capable its built-in search, are real choices on every deployment — so there is no longer a condition reading the value the field itself edits, and nothing to latch. #474's `useResetOnChange(providerId, …)` stays for the `hasInitialized` half; only the latch reset goes with it.
- **The `nativeSearchEnabled` Switch block**, its `providerType !== "Bedrock"` visibility condition, and the `NO_WEB_BACKEND` sentinel (`"none"` is a real value, so Radix needs no stand-in).

Survives, ported onto the one control:

- **The catalog-error note.** A failed catalog fetch still means backend options are missing from what is now the only way to express search at all.
- **The stored-but-uninstalled `<SelectItem>`**, which gains a case: a stale id now sits in the same list as `native` and `none`, so it is gated on the value being neither reserved literal.

`web-search-tool.tsx` and `chat-message.tsx` are untouched: what a search renders as does not depend on how it was selected.

## One addition the plan did not anticipate

**The form also coerces an orphaned `native` when Provider Type or API Mode changes** (`bd2827e`), not only on load. The built-in-search `<SelectItem>` renders only when `providerHasNativeSearch(formData)`, so switching an Anthropic Provider to Bedrock with "built-in search" selected left the select pointing at a value with no matching item — a blank trigger over a stored value the reader cannot see or clear. `handleSelectChange` now coerces it to `None` on that transition, mirroring the load-time coercion decision 2 settled. It is the same failure the load-time coercion exists to prevent, reached through the form instead of the database.

## ADR-0014

Amended, `status` still `proposed` — PR4 flips it. Three Consequences change, all of which the collapse makes false as written:

- **Selection** — "a nullable `webBackend` column" becomes the three-valued `searchSource` column, with a paragraph recording that it replaces a two-field shape that shipped first and why.
- **Turn-time resolution** — restated as the precedence list above.
- **Capability gate** — the gate formula was `nativeSearchEnabled !== false && (providerHasNativeSearch(p) || Boolean(p.webBackend))`; it is now `searchSource !== "none" && (searchSource === "native" ? providerHasNativeSearch(p) : true)`, and `providerHasNativeSearch` gains its second job: deciding whether the built-in option is offered in the selector at all.

The constants amendment #384's approval asked for is **not** here — it is PR4's, with the flip, so there is one PR per Consequences edit.

## Docs

`concepts/providers.mdx` (the Web search section, rewritten around one field), `building-with-platypus/chat.mdx` (the toggle-visibility bullets), `extending/index.mdx` (the selector description, the API field name, and #401's callout about the switch gating a backend — which is now false and goes).

One judgement call: #477 documented three cases in which the PR3 field is visible. All three die — the field is unconditional now — but two of the states it can be **in** survive, so they are kept as content notes rather than dropped with the visibility rules: a selection whose plugin is gone reads `(not installed)`, and a failed catalog load says so and leaves the stored selection alone.

## Also in the diff

Eight comments in files `main` grew after PR3 (`contribution-pipeline.ts`, `contribution-registry.ts` + test, `routes/web-backends.ts`, `server.ts`, `web-backends/index.ts` + `types.ts`, `plugin-sdk/index.ts`) name `provider.webBackend` as the column a backend discriminator is stored in. Renamed to `provider.searchSource` — they are load-bearing comments about where a value comes from, and this PR is what makes them wrong.

## Verification

`pnpm typecheck` clean. `pnpm test` — backend 1819, frontend 193, schemas 126, docs contract 25, all green.

Two pre-existing failures on this machine, both on files byte-identical to `origin/main` and neither touched here:

- `src/runs/sub-agent-lifecycle.test.ts > registers the delegated run for its duration` is **flaky** — two failures and one pass in three consecutive runs of that file alone.
- `pnpm lint` reports 13 `no-unsafe-assignment` / `no-unsafe-return` errors in `src/tools/sub-agent.test.ts` and `src/runs/sub-agent-lifecycle.test.ts`.
- The docs contract test's heading-anchor check fails on a **CRLF checkout**: `headingSlugs` splits on `"\n"`, so every slug keeps a trailing `\r` and all ~60 internal anchors read as broken. With `\r` stripped it passes 25/25, this PR's pages included. Windows-only, pre-existing, not fixed here.

## For the reviewer

**Applying this locally needs the backfill run by hand.** `drizzle-kit push` applies DDL only, not `.sql` files (CLAUDE.md), so a dev database gets `search_source` at its `'native'` default and every Provider that had search off, or a backend selected, silently reads as native until the `UPDATE` from `0051` is run against it. Production takes it through `scripts/migrate.ts` as usual.


willdady (commented):
This reads well — the collapse to a single `searchSource` control is a clear
improvement over the two-column encoding, and the `0050` → `0051` chain and
backfill precedence line up exactly with the old inline `resolveSearchMode`
order. Typecheck is clean and the schemas suite passes. One small thing I'd fix
before merge, one question, and one follow-up.

**The form coercion does get saved** (`apps/frontend/components/provider-form.tsx:600`)

The comment says nothing here saves the coercion back, but `doSubmit` sends
`searchSource` unconditionally at line 819, so it lands on the next save of any
field:

- provider on `apiMode: "chat"` with stored `searchSource: "native"` loads
  coerced to `"none"`
- operator switches API Mode to Responses — the coercion in `handleSelectChange`
  doesn't fire, the value is already `"none"`
- they save, and the provider is now stored as `"none"`

Pre-collapse, `nativeSearchEnabled: true` would have started working the moment
the provider became capable. Either re-derive `"native"` when capability returns
or drop the "without auto-saving" claim — your call which. Worth noting the same
claim is in the `0051` migration comment, so it needs the same treatment there.

**Is `withLegacySearchFields` reachable?** (`packages/schemas/index.ts:1299`)

The docblock describes it as a one-release window for "an API caller still
`PUT`/`POST`ing the pre-collapse shape." I went looking for who that caller
could be and couldn't find one — the provider form is the only writer in the
frontend and always sends a full payload, auth is session-cookie only with no
API-key or bearer path in the middleware, and there's no REST API documented for
operators. If backend and frontend always ship together, I think the mapper and
its two `z.preprocess` wraps (`:1317`, `:1459`) can just go, which shrinks this
PR rather than growing it.

Very possible you're accounting for a caller I can't see from here — if so, I'd
rather know about it than have you delete something load-bearing.

Small aside while you're in there: `searchSource`'s `.default("native")` on the
_update_ schema means the key is always present after parsing, so the update
shape can no longer express "leave this field alone." Nothing can reach that
today, so I'd not hold the PR for it — but it's the invariant the deleted
`"Absent is the pre-existing-row case"` test was pinning, and dropping the
default on update (keeping it on create) would restore it cheaply.

**Follow-up, not this PR:** `"none"` and `"native"` (`:1101`) now share a flat
namespace with web-backend contribution ids, and `registerContributions` only
checks plugin-vs-plugin collisions. It'd take a core backend claiming one of
those ids to bite, so nothing urgent — just flagging that the old
`__platypus_no_web_backend__` sentinel avoided it by construction.

Happy to approve once the form comment/behaviour is settled.

MesoX:
Thanks — you were right on both counts, and the second one shrank the PR. Merged current `main` (`1e46345`) in along the way, which also clears the conflict; the branch is now `11c4207`.

## The form coercion does get saved

Confirmed, exactly as you describe. `doSubmit` sends `searchSource` unconditionally, so the load-time coercion landed on the next save of any field, and a provider whose capability later returned stayed switched off rather than resuming. The claim was wrong in the code comment and in the `0051` migration comment both.

I went with re-deriving rather than dropping the claim, but not by remembering the stored value separately — the stored value is simply never rewritten now. `formData.searchSource` holds exactly what the row holds, and the orphaned selection is **named in the list instead of coerced out of it**: `The provider's built-in search (unavailable here)`. That is the same idiom the stored-but-uninstalled backend already uses, so the select has one rule — show what is stored, say why it does not work — rather than two.

Worth recording why the display-only version did not survive: I wrote it first, coercing the rendered value to `None` while keeping `formData` pristine. A test for "the reader explicitly picks None" then failed, because Radix fires no `onValueChange` when the option picked is the one already showing. The stored `"native"` became unclearable through the form, and the docs sentence promising you could turn it off would have been false. Naming the value keeps it selectable, so picking anything else is a real edit.

Both coercions are gone — the load-time one and the Provider Type / API Mode one from `bd2827e`. A render-time condition covers both routes into the orphaned state, so there is no moment left to latch. Four new tests pin it: the stale selection is named on load, it survives a save that touched only another field, it comes back when API Mode regains native search, and picking `None` still stores `none`.

`concepts/providers.mdx` gains a third bullet in the list that already describes the states this field can be in.

## `withLegacySearchFields` — no, there is no caller, and I have deleted it

I went looking for one on the assumption I had written it for something, and could not find it either. Beyond what you had already spotted:

- The Provider routes take a better-auth session cookie only. `auth.ts` configures `plugins: [admin()]` — no `bearer`, `apiKey` or `jwt` plugin — and no middleware reads an `Authorization` header for them. Every `Authorization` hit in the backend is Platypus acting as an MCP _client_.
- No OpenAPI document anywhere, no `@hono/zod-openapi`, no API-reference page, and no `curl` example against any endpoint in the docs. `@platypuschat/plugin-sdk` is a manifest and type surface with no HTTP in it.
- The two images are built from one commit with one tag in a single job, and both `docker-compose.mdx` and `upgrading.mdx` tell operators to move them together. No `/v1` prefix, no version header, no published client.
- `provider-form.tsx` is the only writer, and it always sends `searchSource` — which trips the mapper's own `"searchSource" in data` guard, so it was already a no-op on every request in production.

The pre-collapse fields did ship publicly (v2.3.0 through v2.7.1), so the window was not guarding a hypothesis — it just has no reachable caller. The one thing I cannot rule out from the repo is someone scripting against the API with a session cookie copied out of a browser; nothing documents or invites that, and I do not think it should hold the shape.

So the mapper and both `z.preprocess` wraps are gone and the two schemas are plain `.pick()`s again. Net **−70 lines** across `packages/schemas`.

## The `.default("native")` on the update schema

Left as-is deliberately, and I would rather flag that than have it read as an oversight.

You are right that it means the update shape cannot express "leave this field alone". But the only caller that could exercise it is the same one that does not exist — the reasoning that just deleted the mapper. Dropping the default on update is also not the one-word change it looks like: the field is `.default(...).transform(...)` and the order matters, so `.pick()` cannot strip it and it needs an `.extend()` override, roughly five lines. Adding five lines to defend an unreachable path in the same PR that deletes a shim for being unreachable seemed like the wrong trade.

One correction to the framing, though: `providerUpdateSchema` has no `.partial()`, so a partial `PUT` already fails on every required field. The regression is specific to `searchSource` — the old `webBackend` was `.nullable().optional()` on update, which is what the deleted test was pinning. Happy to add the override if you would rather have the invariant back than the five lines.

## The flat namespace

Filed as #532 rather than folded in here, so this PR does not take another round. It records that third-party ids are namespaced and so can never collide — it takes a _core_ backend claiming `none` or `native` to bite, at which point `resolveSearchMode` picks the literal and the backend silently never runs.

One thing that came out of sizing it: the tempting two-line fix, pre-seeding the `owners` map, is wrong. That map is also returned as `ContributionOwners` and reports which plugin owns what, so it would surface two fabricated core entries and breaks the `owners.webBackends` assertion in `loader.test.ts`. The honest version is a `validate` hook on `webBackendPoint`, which runs before namespacing and fails boot loudly per ADR-0013.

## Verification

`pnpm typecheck` clean, 7/7. `pnpm test` — backend 1876, frontend 196, schemas 123, plugin-sdk 10, docs contract 24/25.

Two notes on that. Backend is up from 1819 because of what `main` brought in, and the `sub-agent-lifecycle` flake I mentioned earlier passed on every run since the merge. The docs contract failure is the CRLF one from the original description — it fires on `administering/workspaces.mdx`, which this branch does not touch. CI is green on all four checks, which is that same test passing on an LF checkout, so it stays a Windows-only artefact and is still not fixed here.

`main` also added `@ai-sdk/provider`, so this needs a `pnpm install` after pulling or typecheck fails on two test files.

The ADR paragraph recording the back-compat window is in #517, not here, so it comes out as part of that PR along with the status flip.

PR #517, opened 2026-08-15
chore(docs): flip ADR-0014 to accepted and finish the Web-search backend docs

### LINKED ISSUE #329: Feature: pluggable web-search/browse backend routed through the search toggle

MesoX:
# PR4 — Flip ADR-0014 to `accepted`, and finish the Extending docs

Last in the #329 stack, behind #384 (PR1), #393 (PR2), #412 (PR3) and #516 (PR3.5).

Closes #329.

> [!NOTE]
> **Maintainer edit:** #516 has since merged and `main` has been merged back in here, so the diff is now what this PR is actually about — 2 files for the ADR, 5 for the docs, plus review fixes to `CONTEXT.md` and the pages listed in the review below. The earlier "do not merge before #516" warning no longer applies.

## What this PR is

**The ADR flip.** ADR-0014 was the only `proposed` ADR in the set; the other seventeen are `accepted`. The Extension point has now shipped end to end — SDK contract, registry and loader (#384), turn-time resolution and the schema column (#393), the catalog route, Provider selector and Sources rendering (#412), and the `searchSource` collapse (#516) — so the design of record matches production and there is nothing left to hold it open.

The flip also corrects one stale line. The summary paragraph still described selection as "a nullable `webBackend` column" while the Consequences directly below it record the three-valued `searchSource` column that replaced it. Accepting an ADR whose opening contradicts its own Consequences would pin the wrong design as the record, so the summary now names `searchSource`. No other body change — the Consequences amendments for the core-owned constants and the collapse landed earlier on this branch.

**The docs.** `apps/docs/content/extending/` gained a Web-search backends page earlier in the branch. The final commit here closes the gaps a newcomer actually hits, found by reading the published pages end to end as someone building a third-party SearXNG backend rather than by diffing against the code.

The substantive one: **the Web-search backend page was the only contribution page with no link to `@platypuschat/plugin-sdk`.** Both siblings — Tool sets and Sandbox backends — open with one and send the reader to the package for the shapes the pages deliberately don't paste ("a signature pasted onto a docs page drifts silently"). Web-search backends inherited the structure but not the link, so an author following the docs reached the caps, the executor rules and the egress guard with nowhere to read the executor shapes from. The worked example already exists in the SDK README; nothing pointed at it.

The rest are small and factual:

- `searchSource` and the Provider-form path are now named on the contribution fields, folding in a duplicate prose section from the Extending hub page that no other Extension point had.
- The deploy-time config delivery list named Tool sets and Sandbox backends only, while the Web-search page delegated its whole credentials story to that list.
- The quickstart's "Where to next" omitted the Extension point and still called Sandbox backends "the other" one.
- **Organization settings → Plugins** has listed Web-search backends since #384, but `self-hosting/configuration.mdx` still documented it as Tool sets and Sandbox backends only — the first thing a plugin author checks to confirm their contribution loaded.

## What I deliberately did not do

I first proposed three further additions and then checked them against the rest of the docs rather than against my own reasoning, which killed all three. A code block on the Web-search page looked obviously right until I noticed **`sandbox-backends.mdx` — the closest sibling, same contribution-with-a-factory shape — has zero code fences**; I had benchmarked against `tool-sets.mdx`, where literal tool objects are the subject matter. A `configSchema` example would have been the only manifest declaration in code anywhere outside the quickstart. And a checkpoint chain would have imported a section type that exists on exactly one page, the quickstart, where it belongs. The gap was real in all three cases; the fix was one missing link, not three new sections.

## Testing

`pnpm --filter docs test` — 25/25 passing, including the internal-link and heading-anchor contract.

One thing worth flagging separately, unrelated to this change: **that suite cannot pass on a Windows checkout with `core.autocrlf=true`.** `headingSlugs` matches lines with `/^#{1,6}\s+(.*)$/`, and on a CRLF line `.` does not match `\r` while `$` will not anchor before it, so **zero** headings are collected and every anchor link reports as unresolvable — about 40 false violations across the docs. CI is Linux/LF so it is green there. I confirmed it fails identically on a clean tree, then normalised line endings locally to verify this PR. It is a one-line fix (`(.*?)\r?$`, or strip `\r` before matching) and I have left it out of this PR rather than widen a docs PR into the test harness — happy to send it separately if you want it.


MesoX:
I have also surfaced all other questions/potential PRs as issues so we can close the #329 (#519, #520, #521, #522, #523, #524, #525).


Thanks, this really is something that brings Platypus to world class tool for us!

willdady (approved):
Approving — and thank you, @MesoX. #329 landed as five PRs over its life (#384 the Extension point, #393 turn-time resolution, #412 the selector and Sources rendering, #516 the `searchSource` collapse, and this one), plus ADR-0014 as the design of record. That is a lot of sustained work held to a consistent shape, and the shape is what makes it reviewable.

Two things I want to call out specifically.

The **"What I deliberately did not do"** section is the most useful part of this description. Proposing three additions, checking them against the rest of the docs rather than against your own reasoning, and killing all three — including noticing that `sandbox-backends.mdx` has zero code fences and that you had been benchmarking against the wrong sibling — is the judgement the docs-audit voice guide is trying to encode and mostly cannot. The result is one missing SDK link fixed rather than three new sections nobody asked for.

The **CRLF finding is real** and I confirmed it: `headingSlugs` splits on `\n` and matches `/^#{1,6}\s+(.*)$/`, so on a CRLF checkout the trailing `\r` defeats `$` and zero headings are collected — every anchor then reports unresolvable. Please do send `(.*?)\r?$` as its own PR; you were right to keep it out of a docs change, and it should not sit in your local tree.

**Housekeeping**, since our timezones do not overlap and I did not want to hold this: I pushed two commits to your branch rather than send it back.

- A **merge of `main`** now that #516 has landed. The conflicts were all the squash-vs-branch kind — I took `main`'s side for `provider-form.tsx` and its test (it carries #516 plus the #534 cleanup, and this PR changes neither), and your side for the ADR and the docs pages. The diff is back to the files this PR is actually about.
- A **review-fix commit**. Two were real: `CONTEXT.md` still described a Web-search backend as "a planned addition" while this PR accepts the ADR that shipped it, and ADR-0014's observability bullet claimed the content-cap warn was folded into the per-call record when only the egress block was. The rest are precision: the page-reader warning was written out on three pages where `VOICE.md` asks for it once, so Chat now links to Providers; the robots.txt contrast was absolute where `config.ignoreRobotsTxt` makes it a default; and `read_url`'s content type is optional in the SDK.

Every number, field name and enum on the new page checked out against the shipped module — caps, timeouts, the boot-time refusal, the egress ranges, the capability predicate. Nothing in the ADR describes something that did not ship, which is the bar for flipping it to `accepted`.

I have also added `Closes #329` to the description so the issue closes with the merge.

PR #518, opened 2026-08-16
chore(ci): disable Turborepo telemetry in CI workflows

### LINKED ISSUE #506: chore(ci): disable Turborepo telemetry in CI
## Problem

Turborepo collects anonymous usage telemetry by default, and our CI runners are sending it on every run. Any job that invokes `turbo` phones home to Vercel:

- `.github/workflows/ci.yml` — `pnpm test`, `pnpm typecheck`
- `.github/workflows/lint.yml` — `pnpm lint` (also called by `build-and-push.yml` and `deploy-docs.yml`)

We get nothing from it, and CI shouldn't be making unnecessary outbound calls on our behalf.

## Proposal

Set `TURBO_TELEMETRY_DISABLED=1` as a workflow-level `env:` in the workflows that run turbo. The env var is the right lever for CI — `npx turbo telemetry disable` writes to a machine-local config file, which a fresh runner never has.

Open question for triage: set it per-workflow, or once at the repo/org level (Settings → Variables) so any future workflow inherits it? A repo variable is fewer edits but invisible in the diff; per-workflow `env:` is explicit and reviewable.

## Reference

https://turborepo.dev/docs/telemetry#disable

snowyukitty:
### Problem
Turborepo collects anonymous usage telemetry by default, sending outbound telemetry calls on each CI invocation that runs `turbo` (such as `test`, `typecheck`, and `lint`).

### Solution
Set `TURBO_TELEMETRY_DISABLED: "1"` at the top-level `env:` scope in `.github/workflows/ci.yml` and `.github/workflows/lint.yml` to prevent unsolicited outbound network telemetry from CI runners.

### Reference
- https://turborepo.dev/docs/telemetry#disable

Closes #506

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,9 @@ on:
 permissions:
   contents: read
 
+env:
+  TURBO_TELEMETRY_DISABLED: "1"
+
 # Cancel superseded runs for the same ref (e.g. a new push to a PR branch).
 concurrency:
   group: ci-${{ github.ref }}
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,6 +9,9 @@ on:
 permissions:
   contents: read
 
+env:
+  TURBO_TELEMETRY_DISABLED: "1"
+
 jobs:
   lint:
     runs-on: ubuntu-latest


willdady (approved):
@snowyukitty thanks for the contribution!

PR #535, opened 2026-08-16
feat(backend): collapse the three run drivers into runs/drive

### LINKED ISSUE #496: Collapse the three run drivers into runs/drive
From the 2026-08-15 backend architecture review.

## Problem

#487 unified the run *lifecycle* (all paths call `startRun`) but not the *driving*: three code paths each re-derive "consume the stream, keep the latest message, decide succeeded/failed/cancelled, record a ceiling cutoff":

- `apps/backend/src/runs/agent-runner.ts:281-345` (`stream`) and `:352-447` (`generate`)
- `apps/backend/src/tools/sub-agent.ts:299-480` (delegate `execute`)

Pairwise duplicates: `startRun` (`agent-runner.ts:174` / `sub-agent.ts:299`); `buildModelInvocation` (`:240` / `:358`); `toUIMessageStream` + `createMessageMetadata` + `formatStreamError` (`:295-313` / `:379-382`); the `readUIMessageStream` snapshot loop (`:322-343` / `:390-402`); the terminal-status decision (`:310`, `:440-444` / `:428-433`, `:462`, `:479`). Output-ceiling truncation is recorded three different ways (`agent-runner.ts:383-385`, `sub-agent.ts:454-455`, `runs/message-metadata.ts:102-104`). Consequence already visible: no-progress detection is wired only to `generate` (`agent-runner.ts:231`, `:368`), so a Trigger's delegated runs — equally unattended — silently get none.

## Solution

`runs/drive.ts` owns the model call inside a registered run:

- `driveStreamed(plan, run, { onSnapshot }) → { messages, latest, failure, truncated }`
- `driveOnce(plan, run) → { text, stats }`

`AgentRunner.stream` keeps only the tee + `Response`; `generate` keeps only its return shape; the delegate tool keeps only the activity projection and the throw-on-failure rule. Terminal status, ceiling recording and stop conditions are decided once, inside the seam.

## Wins

- Terminal-status rule lives in one place
- Ceiling cutoff recorded once, three readers
- No-progress reaches delegated unattended runs
- Stream consumption testable without a runner
- New entry points inherit bounds by construction

## ADR

None.

## Dependency

Consumes the `RunPlan` produced by `runs/agent-plan` (#495) — land that first.

willdady:
Closes #496.

## What

Extracts the model drill from the three places that re-derived it — the Chat `stream` and headless `generate` paths in `AgentRunner`, and the delegate tool's `execute` — into `apps/backend/src/runs/drive.ts`.

The seam now owns what used to be copied three ways:
- building the model invocation (step ceiling and the no-progress stop condition included),
- the terminal-status decision (succeeded / failed / cancelled),
- recording the output-ceiling cutoff on the run's stats,
- ending the run.

`driveStreamed` serves `streamText` runs (a Chat turn tees a client stream; a delegated sub-Agent folds the snapshots into its activity log) and `driveOnce` serves `generateText` runs (Triggers).

A delegated sub-Agent run is unattended, so it now drives under the same no-progress condition a Trigger has — a stuck model that re-issues the same call for the same result is aborted instead of running to the step ceiling (previously wired to `generate` only, so Trigger-delegated runs silently got none).

## Callers keep only their shape

- `AgentRunner.stream` — the tee + `Response`
- `AgentRunner.generate` — `{ text, stats }`
- delegate tool — the activity projection + the throw-on-failure rule

## Tests

- New `runs/drive.test.ts` at the seam (real AI SDK pipeline, mock model)
- Full backend suite, typecheck, ESLint and Prettier all green (1884 tests)

## ADR
None.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -34,6 +34,10 @@ A configurable preset that pins a Provider, model, Instructions, generation para
 **Sub-Agent**:
 An Agent referenced by a parent Agent and exposed to it as a delegate Tool. Invoking it starts a run in its own right — bounded by the same per-step and per-run timeouts the parent turn was started under, and cancelled when the parent is — but never a Chat: nothing about a delegated run is persisted.
 
+**Drive**:
+Running a resolved turn's model loop inside a registered run, through to a terminal status. Takes one of three shapes according to who is waiting on it: an interactive **Chat turn**, a delegated **Sub-Agent**, or a headless Trigger. Whichever shape, the drive owns the model invocation and its stop conditions, the succeeded / failed / cancelled decision, and recording an **Output ceiling** cutoff; the entry point that asked for it keeps only the result shape it hands back to its own caller.
+_Avoid_: driver, run executor, generation loop.
+
 **Instructions**:
 The free-text behaviour brief a User writes on an Agent — or on a Chat with no Agent. One input to the System prompt rather than the whole of it: it renders as the first fragment and cannot suppress the Platypus-owned fragments that follow.
 _Avoid_: system prompt (that names the composed artefact), prompt, persona.
--- a/apps/backend/src/runs/agent-runner.test.ts
+++ b/apps/backend/src/runs/agent-runner.test.ts
@@ -1101,7 +1101,13 @@ describe("AgentRunner.stream — success & interruption", () => {
       ],
     });
     queue.end();
-    await tick();
+    // The run is finished by the drive once the background snapshot branch
+    // drains (and teardown disposes the turn, pushing the racing snapshot).
+    // Await that drain deterministically before reading the terminal sink write.
+    await vi.waitFor(
+      () => expect(runRegistry.has("s-late-snapshot")).toBe(false),
+      { timeout: 2_000 },
+    );
 
     const finish = sink.events.at(-1) as Extract<
       LifecycleEvent,
--- a/apps/backend/src/runs/agent-runner.ts
+++ b/apps/backend/src/runs/agent-runner.ts
@@ -261,9 +261,10 @@ export class AgentRunner {
       agentId: state.turn!.resolved.agentId,
       generateMessageId: createIdGenerator({ prefix: "msg", size: 16 }),
       toolDurations,
-      // Interactive runs leave no-progress off — a human can stop those.
-      stopSnapshotsAfterFinal: true,
-      returnResponse: true,
+      // An interactive Chat turn: no-progress left off (a human can stop
+      // those), a stream error rendered inline rather than failing the run,
+      // and a client stream split off.
+      mode: "chat",
       onToolExecutionEnd: ({ toolCall, toolExecutionMs }) => {
         toolDurations.set(toolCall.toolCallId, toolExecutionMs);
       },
--- a/apps/backend/src/runs/agent-runner.ts
+++ b/apps/backend/src/runs/agent-runner.ts
@@ -2,13 +2,8 @@ import {
   convertToModelMessages,
   createIdGenerator,
   createUIMessageStreamResponse,
-  generateText,
-  readUIMessageStream,
-  streamText,
 } from "ai";
-import { formatStreamError, isTruncatedByTokenLimit } from "./stream-error.ts";
-import { createMessageMetadata } from "./message-metadata.ts";
-import { applyToolDurations } from "./tool-durations.ts";
+import type { ModelMessage } from "ai";
 import {
   prepareChatTurn,
   validateTurnAttachments,
@@ -20,13 +15,7 @@ import { actorUserId, type WorkspaceScope } from "../scope.ts";
 import type { PlatypusUIMessage } from "../types.ts";
 import { runRegistry, type RunTimeouts } from "./run-registry.ts";
 import { startRun } from "./run-lifecycle.ts";
-import { buildModelInvocation } from "./run-plan.ts";
-import { computeStats } from "./run-stats.ts";
-import {
-  createNoProgressDetector,
-  NoProgressError,
-  type NoProgressDetector,
-} from "./no-progress.ts";
+import { driveOnce, driveStreamed } from "./drive.ts";
 import type {
   ResolvedRunPlan,
   RunId,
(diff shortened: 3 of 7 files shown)


willdady:
## Code review — two axes

Reviewed `origin/main...origin/feature/496-collapse-run-drivers` (2 commits, 6 files, +840/−251) against the repo's documented standards and against issue #496. The two axes ran independently and are reported separately, unmerged.

---

## Standards

**Documented standards — no hard violations found.**

Commits (`feat(backend):`, `chore(backend):`) and branch `feature/496-…` match CLAUDE.md/CONTRIBUTING. No parameter properties. Prettier clean (verified with `npx prettier --check`). The diff touches only `apps/backend/src/runs/**` and `src/tools/**` — it intersects no row of the CLAUDE.md docs table (no `.env.example`, no `packages/schemas` limit/enum, no `plugins/**`, no frontend label), so no `apps/docs` edit is owed. ADR-0016 is honoured: the Sub-Agent system-prompt exception moved intact into the `plan` handed to `driveStreamed`.

Soft: "drive" is now a load-bearing term (module, two functions, commit subject) but isn't in `CONTEXT.md`; `docs/agents/domain.md` says a missing glossary term is a signal to note for `/domain-modeling`. `CONTEXT.md` also lists "driver" as a term to avoid (for Gateway adapters) — different context, but the commit subject reuses the word.

**Baseline smells (all judgement calls)**

1. **Duplicated Code — `drive.ts`.** The docblock claims one terminal rule, but `endStreamedRun` and the generator's `finally` re-derive it independently: `const trip = noProgress?.tripped() ?? null; … new NoProgressError(trip.toolName, trip.count).message … run.handle.signal.aborted ? …`. Recorded status and returned `failure` can drift.
2. **Duplicated Code — `driveStreamed`/`driveOnce`.** Verbatim in both: `unattended ? createNoProgressDetector() : null` and `...(prompt !== undefined ? { prompt } : { messages: modelMessages ?? [] })`.
3. **Data Clumps / flag params — `StreamedDriveOptions`.** Four booleans, each set by exactly one caller: `{stopSnapshotsAfterFinal, returnResponse}` = Chat only; `{unattended, failOnStreamError}` = delegate only. The bag encodes two modes rather than one seam.
4. **Primitive Obsession.** `prompt?: string` / `modelMessages?: ModelMessage[]` are "mutually exclusive" in prose only; neither-supplied silently becomes `messages: []`. A union makes that illegal state unrepresentable.
5. **Middle Man / duplicated terminal decision — `sub-agent.ts`.** `await run.finish(cancelled ? "cancelled" : "failed", …)` with a comment conceding "when the drive already ended the run … this is a no-op". Live only for the setup path; the `drive: … | null` + `setupError` + `drive!.done` dance exists just to serve it.
6. **Duplicated Code — `agent-runner.stream`.** Its catch logs `"Server-side UI stream consumer error"`, identical to what `drive.ts` already emits for the same condition; the drive swallows the error, so the branch is near-dead.
7. **Mysterious/orphan comment.** In `stream()`, `// Interactive runs leave no-progress off — a human can stop those.` sits over `stopSnapshotsAfterFinal: true`, describing an option not passed. The "never overwrites the folded final with a duration-less snapshot" sentence is repeated twice in one function.
8. **Feature Envy (minor).** `run.setStats({ ...run.stats, truncatedByTokenLimit: true })` — read-modify-write of `RunLifecycle`'s stats from outside; wants `run.markTruncated()`. Also `handoverMessages.some(...) ?? false` — the `??` is dead. Three `state.turn!` / `drive.response!` assertions leak that `setup`'s return type doesn't promise what both callers need.

---

## Spec (issue #496)

### (c) Implemented but wrong

**1. Chat truncation flag scans the whole conversation → false positives.** `drive.ts` onFinish:

```js
truncated = handoverMessages.some(
  (m) => m.metadata?.truncatedByTokenLimit === true,
);
```

`handoverMessages` is `toUIMessageStream`'s output, which includes `originalMessages` (= `input.messages`, the persisted chat history). A turn _earlier_ in the chat that was truncated carries that metadata persistently, so every subsequent clean turn now sets `run.setStats({truncatedByTokenLimit: true})`. Old code (`message-metadata.ts:102-104`) only stamped the _terminal finish part of this stream_. Spec: "Ceiling cutoff recorded once, three readers" — recorded once, but from the wrong scope.

**2. Delegate abandonment now records `succeeded`, not `cancelled`.** `sub-agent.ts` still has `finally { await run.finish("cancelled") }` — "Reached without an outcome only when nobody consumed the result." But when the parent abandons the generator, for-await cleanup calls `.return()` on `drive.snapshots` first, whose `finally` runs `endStreamedRun`; the signal isn't aborted, so it finishes **succeeded**. `finish` is once-only, so the outer `"cancelled"` is now a no-op. The pre-change comment (`sub-agent.ts:296` on main) explicitly guaranteed the opposite. Not covered by `drive.test.ts` (only "stopped mid-stream" via signal is tested).

**3. Chat response can close before the run is persisted.** `run.finish` moved out of the SDK-awaited `toUIMessageStream.onFinish` into the snapshot generator's `finally`. The two tee branches are independent, so the client body can now complete before `sink.onFinish` writes. Previously ordered.

**4. Lost log.** The chat snapshot branch's `onError` used to log `"Snapshot stream parse error"`. It now assigns `failure`, which a `failOnStreamError: false` drive discards — the parse error vanishes.

### (a) Missing / partial

**Spec:** "the delegate tool keeps only the activity projection and the throw-on-failure rule." `sub-agent.ts` still calls `run.statusFromSignal()` and re-decides `cancelled ? "cancelled" : "failed"` — the exact terminal-status re-derivation the spec moved into the seam.

**Spec:** "Ceiling cutoff recorded once." Still derived two ways: metadata scan (`driveStreamed`) vs `isTruncatedByTokenLimit(result.finishReason)` (`driveOnce`), plus `message-metadata.ts` unchanged.

### (b) Scope creep / deviation

**Spec:** "`AgentRunner.stream` keeps only the tee + `Response`." The tee was moved _into_ `drive.ts` behind `returnResponse` — the opposite split. Combined with `stopSnapshotsAfterFinal`, `failOnStreamError`, `unattended`, the spec's `driveStreamed(plan, run, {onSnapshot})` became a 13-field option bag with four caller-mode booleans — the seam is configured by its callers rather than owning the rules.

---

**Summary** — Standards: 0 hard violations, 8 judgement calls + 1 soft glossary note; worst is the duplicated terminal-status derivation inside `drive.ts` itself (#1), which undercuts the change's own premise. Spec: 7 findings (4 wrong, 2 partial, 1 deviation); worst is the truncation flag scanning persisted chat history, silently marking clean turns as ceiling-truncated.

PR #580, opened 2026-08-19
chore(frontend): delete the unreachable blob-to-data conversion in the chat surface

### LINKED ISSUE #574: chore(frontend): delete the unreachable blob-to-data conversion in the chat surface
## What to build

Delete a block of unreachable code in the chat surface. Before sending a message, it walks the attached files converting any blob URL into a data URL — but the composer has already done exactly that conversion before handing the files over, so the guard that skips already-converted files is always true and the loop never does anything.

The change is a deletion, and the behaviour is provably unchanged.

**Out of scope:** the underlying cause is that the type the composer hands over does not record that its URLs are already data URLs, so the caller defends against a state the producer had already excluded. Do not change that type here — mention it in the PR description and leave it. This ticket deletes the dead loop and nothing else.

## Acceptance criteria

- [ ] The dead conversion is deleted from the chat surface
- [ ] Sending a message with an image attachment still works
- [ ] Sending a message with a non-image file attachment still works
- [ ] No type or signature outside the deleted block is modified

## Blocked by

None — can start immediately.

mehmetturuncx:
Closes #574

Delete a block of unreachable code in the chat surface. Before sending a message, it used to walk the attached files converting any blob URL into a data URL. However, the composer already does exactly that conversion before handing the files over, so the guard that skips already-converted files is always true and the loop never does anything.

**Note:** As mentioned in the issue, the underlying cause is that the type the composer hands over does not record that its URLs are already data URLs, so the caller defends against a state the producer had already excluded. This PR does not change that type, it only deletes the dead loop.

### DIFF
--- a/apps/frontend/components/chat.tsx
+++ b/apps/frontend/components/chat.tsx
@@ -520,30 +520,12 @@ export const Chat = ({
       return;
     }
 
-    // Convert blob: URLs to data: URLs so the backend can access the file
-    // content. blob: URLs are browser-only references that can't be resolved
-    // server-side.
-    const files = await Promise.all(
-      (message.files ?? []).map(async (file) => {
-        if (!file.url.startsWith("blob:")) return file;
-        const res = await fetch(file.url);
-        const blob = await res.blob();
-        const dataUrl = await new Promise<string>((resolve, reject) => {
-          const reader = new FileReader();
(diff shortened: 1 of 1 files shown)


willdady (approved):
Thanks for the clean deletion — verified the loop was genuinely dead code and the PR stays in scope. LGTM!

Issue #627, opened 2026-08-21
[Web-search backend] Two of the three no-search warns name the plugin but not the Provider


MesoX:
A Chat turn that asks for search and gets none now tells the User so (#522). The Operator side of that pair is three warn lines, and only one of them can be traced back to a row anyone can edit.

`resolveSearchTools` (`apps/backend/src/services/chat-execution.ts`) logs `{ orgId, workspaceId, providerId, searchSource }` when a stored `searchSource` names a backend with no registration. The other two — `createExecutors` threw or outran its `timeoutMs`, and a factory that returned no `web_search` executor — are raised inside `buildTurnTools` (`apps/backend/src/web-backends/index.ts`) and log `{ plugin, backend }` plus the `cause`. #522 adds `orgId` and `workspaceId` to those two, which were already in scope on the turn's `WebBackendContext` and simply unlogged. That is as far as that PR can go: `WebBackendContext` is `{ orgId, workspaceId, userId }` and carries no `providerId`.

## Why this matters now

Before #522 these two degradations were invisible to everyone except whoever read the log. Now the User sees **Search was unavailable — this reply was written without it.** under the reply, and the docs tell them to take it to their Operator. The Operator arrives at a line that names a plugin.

The missing field is the one the registry-miss warn already carries, for a reason recorded in its own comment: an org-scoped Shared Provider (ADR-0007) is one row serving many Workspaces, so the Workspace names a symptom while the fix is an edit to a Provider the line cannot identify. That argument applies with more force to the factory-failure case, because it is the common one in practice — a self-hosted SearXNG or browser service that is down, slow or restarting — and because "which Provider points at this backend" is not answerable from a plugin id once more than one Provider names the same backend.

It also leaves one of #522's acceptance criteria unsatisfiable on its literal wording: "All three server-side warns still fire with their existing distinct messages and still carry the Provider id." Two of them never carried it, and #522 ships on the reading that the criterion means the warns are unchanged and the one that has `providerId` keeps it.

Honest sizing: small, and diagnosability-only. Nothing a User sees changes.

## Recommended approach

**Do not widen `WebBackendContext`.** ADR-0014 defines it as the runtime scope handed to a *plugin's* executor factory, deliberately mirrored on `SandboxContext` and deliberately minimal. A Provider row id is a core concern no backend author has any business receiving, and adding it would land in the published plugin SDK (`packages/plugin-sdk`) permanently.

[...]

Issue #624, opened 2026-08-21
Search toggle stays on when switching to an Agent or Provider that cannot search


willdady:
> _This was generated by AI during triage._

Found while triaging #522. Not part of that fix — #522 reports an unavailable
search, it does not change the toggle's lifecycle — but #522 will make this more
visible to Users, so it is worth closing separately.

## Current behaviour

The Chat search toggle is reset by one effect keyed on the direct-selection
fields:

```ts
useResetOnChange(`${modelId}:${providerId}`, () => setSearch(false));
```

Selection is mutually exclusive. Picking an Agent sets `agentId` and clears
`providerId` and `modelId`; picking a Provider and model sets those two and
clears `agentId`. So while **any** Agent is selected, both watched values are
empty strings and the watched key is the literal `":"`.

The bug is not a missing `agentId` in that key. It is that the key carries no
information at all while an Agent is selected — every Agent produces the same
key, so the reset is structurally unable to fire between Agents.

| Switch from            | to                     | key before | key after   | resets? |
| ---------------------- | ---------------------- | ---------- | ----------- | ------- |
| model on Provider `P1` | Agent "Researcher"     | `gpt-4:P1` | `:`         | yes     |
| Agent "Researcher"     | Agent "Drafter"        | `:`        | `:`         | **no**  |
| Agent "Researcher"     | model on Provider `P1` | `:`        | `gpt-4:P1`  | yes     |
| model on `P1`          | model on `P2`          | `gpt-4:P1` | `claude:P2` | yes     |

`search` is sent on every turn either way — the request body is
`{ agentId, search }` when an Agent is selected, and
`{ providerId, modelId, …, search }` otherwise.

### What a User sees

Take a Workspace with three Agents: **Researcher** on a vLLM Provider whose Web
search is a SearXNG backend; **Drafter** on an OpenAI Provider using the vendor's
own search; **Checker** on a Bedrock Provider, which has no search at all.

Select Researcher, turn the globe on, ask some research questions. Search runs,
correctly. Switch to Drafter to write the summary: no reset fires, the globe is
still drawn and **still lit**, and every Drafter turn now searches the web
although the User never turned search on for Drafter.

Worse, switch instead to Checker. Bedrock cannot search, so the capability
predicate is false and **the globe is not rendered**. But `search` is still
`true` and is still sent; Turn resolution resolves it to nothing, so no search
runs. There is now a setting that is on with no control on screen to turn it off.
[...]

Issue #478, opened 2026-08-15
Kanban: one module behind the HTTP routes and the Tool set (Tool path skips validation the route enforces)


willdady:
## Problem

The Kanban board's behaviour is implemented twice — once in `apps/backend/src/routes/kanban.ts` (1258L) for the HTTP surface, once in `apps/backend/src/tools/kanban.ts` (1126L) for the Agent Tool set. Only the pure arithmetic is shared (`utils/kanban-positioning.ts`, `utils/kanban-labels.ts`). The copies have already diverged, so this is a live correctness bug, not just duplication.

### Shipped divergence (the bug)

`routes/kanban.ts` has two module-private validators that only the HTTP path calls:

- `validateAssignees` (`routes/kanban.ts:51-90`) — assignees must be org members, super admins, or Agents visible in this Workspace (via `listScopedByIds`). Called at `routes/kanban.ts:767-775` and `:839-847`.
- `getBoardLabelIds` + `filterKnownLabelIds` (`routes/kanban.ts:134-142`) — label ids must exist on the board. Used at `:757-763`, `:833-836`.

`tools/kanban.ts` calls **neither** and writes raw values:

- `tools/kanban.ts:357` — `updateData.labelIds = labelIds` unvalidated
- `tools/kanban.ts:422-423` — `labelIds: labelIds ?? [], assignees: assignees ?? []` unvalidated
- `tools/kanban.ts:1043-1057` — bulk update, same

Consequence: an Agent can persist label ids that are not on the board and assignees who are not org members or visible Agents. The HTTP surface rejects both.

### Duplication evidence

- `moveCard`: `routes/kanban.ts:903-964` vs `tools/kanban.ts:476-537` — ~52 lines identical (same variable names, same rebalance loop, same `dispatchEvent(\"card.updated\", ...)`). Only deltas: `lastEditedByUserId: user.id` vs `lastEditedByAgentId: agentId`, and the error envelope.
- A third copy of the append+rebalance lives inline in `duplicateCard` (`tools/kanban.ts:789-830`).
- The append-to-end idiom `(max(position) ?? 0) + 1.0` has five copies: `routes/kanban.ts:617-621`, `:779-783`; `tools/kanban.ts:407-411`, `:794-797`, `:1023-1040`.
[...]

Issue #312, opened 2026-07-09
chore(backend): migrate PLATYPUS_SANDBOX_DOCKER_ALLOWED_NETWORKS to @platypus/docker plugin config


willdady:
## Context

Surfaced while documenting the SSH sandbox backend (#287). Under the plugin model (ADR-0013), deploy-time, operator-owned plugin settings are expected to flow through `PLATYPUS_PLUGIN_CONFIG`, keyed by plugin name. The Docker sandbox backend predates that mechanism and still reads its operator allowlist directly from a dedicated env var.

## Current state

- `apps/backend/src/plugins/docker/backend.ts:45` reads `process.env.PLATYPUS_SANDBOX_DOCKER_ALLOWED_NETWORKS` directly.
- `apps/backend/src/plugins/docker/index.ts` declares **no** plugin-level `configSchema`/`credentialsSchema` — the `configSchema` there is the *per-Workspace* Sandbox config, not `PLATYPUS_PLUGIN_CONFIG`.
- `apps/docs/content/self-hosting/configuration.mdx` already notes: "`@platypus/docker` keeps taking its per-Workspace config as before and needs no entry here."

So the allowlist is a one-off env var sitting outside the standard plugin-config surface that every other operator-owned plugin setting uses.

## Proposal

Move the operator network allowlist onto `@platypus/docker`'s plugin-level `configSchema`, consumed via `PLATYPUS_PLUGIN_CONFIG`, e.g.:

```jsonc
{ "@platypus/docker": { "config": { "allowedNetworks": ["shared", "tools"] } } }
```

This makes the Docker backend consistent with the plugin model and keeps all deploy-time plugin config auditable in one place.

## Design decisions for triage

- **Deprecation vs breaking change** — keep reading `PLATYPUS_SANDBOX_DOCKER_ALLOWED_NETWORKS` as a fallback with a deprecation warning for a release or two, or cut over hard?
- The route/allowlist-validation path (`backend.ts:82` references the env var name in its error message) needs updating to point at the new config source.
- Docs to update once landed: `reference/backend-configuration.mdx` (Docker sandbox table), `self-hosting/configuration.mdx` and `self-hosting/sandbox.mdx` (allowlist narrative).

## Out of scope

The SSH backend (#287) — it has no deployment env vars and is unaffected.

willdady:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** enhancement
[...]

PR #628, opened 2026-08-21
fix(backend,frontend): report a Chat turn that served no search tools

### LINKED ISSUE #522: [Web-search backend] A chat turn whose backend is no longer installed answers with no search and tells nobody
An Operator drops a search plugin from `PLATYPUS_PLUGINS` while a Provider still names its backend. The Provider form already handles its half: the stored id is listed with a "(not installed)" suffix once the catalog has loaded (`apps/frontend/components/provider-form.tsx:1357-1373`). Chat does not: the globe still appears, the user turns it on, asks a question, and gets an answer produced with no search at all, with nothing on screen saying so. The only record is a server-side warn line.

This is written against #516 and #517, which land before any work here would start — #516 collapses `nativeSearchEnabled` + `webBackend` into `searchSource`, and #517 carries the ADR wording quoted below. The gap itself predates both: on `main` today the same expression ends in `Boolean(webBackend)`.

## Why this might matter

The chat gate trusts the stored value alone. `canSearch` (`apps/frontend/components/chat.tsx:436-441`) tests capability only for `"native"`; for a backend id it is `Boolean(resolvedProvider.searchSource)`, with no check that anything is still registered, and that expression renders the globe (`chat.tsx:641`). Server-side, `resolveSearchMode` (`apps/backend/src/services/chat-execution.ts:447-459`) returns `{ kind: "backend" }` for any non-`"none"`/non-`"native"` value, and `resolveSearchTools` (`chat-execution.ts:469-501`) finds no registration, logs `{ orgId, workspaceId, providerId, searchSource }` with `"provider.searchSource references an unregistered web backend; serving no search tools this turn"`, and returns `{}`. That degradation is deliberate — ADR-0014's Selection bullet is explicit that the stored value is "**not** validated against the registry there: the registry is a backend-runtime concern, and a stale id degrades to no search tools plus a warn-log rather than blocking the form." Nothing below touches the form.

Honest sizing: this only happens after an Operator removes an installed plugin while a Provider still points at it — a misconfiguration-recovery path, not an everyday one. It is still worth closing, because a silently unsearched answer is the worst failure mode this feature has: the user cannot tell it happened, and neither can anyone reading the transcript later.

## What it would fix

- The globe stops being offered, or stops being silent, on a Provider whose backend is gone.
- A user gets either no search affordance or a visible "no search ran" marker, not a confident stale answer.
- The condition becomes diagnosable by the person affected, not only by whoever reads the logs.

## Recommended approach

Two shapes, and we lean toward the second.

**A — hide the toggle (the shape originally deferred).** `chat.tsx` fetches `GET /organizations/:orgId/web-backends` (`apps/backend/src/routes/web-backends.ts:30`, mounted at `server.ts:190` behind `requireAuth` + `requireOrgAccess()`, so every org member can already read it), caches it, and adds a membership test to `canSearch`. That puts the frontend in the business of mirroring backend registry state — what the ADR bullet declines to do at the form — and it fails the wrong way: a failed catalog request hides the toggle where search works fine, unless the "unknown" state gets the care `webBackendsKnown` gives it in the form (`provider-form.tsx:566-567`).

**B — report the dead turn instead of hiding the control.** The warn at `chat-execution.ts:489-497` already knows the exact fact the user needs. Carrying it into the turn's message metadata and rendering a one-line notice reuses the same _channel_ `truncatedByTokenLimit` uses (`apps/backend/src/types.ts:17-24`, `CutShortNotice` at `apps/frontend/components/cut-short-notice.tsx:13`) — but not the same trigger point. `truncatedByTokenLimit` is set from a `finish` part at `message-metadata.ts:102-104`, after the model call completes; the dead-search fact is known much earlier, at tool-setup time inside `resolveSearchTools`, before the model is ever called. So this needs a new emission point around `chat-execution.ts:489-497` itself, not a copy of the finish-handler logic. That is otherwise the fix #420 got for the same class of problem: a turn that went wrong visibly only in the operator's log. No extra fetch, no cache, no registry mirror; it names the wasted turn rather than preventing it.

The two are not exclusive, but B is the one that's necessary regardless: A only narrows the window, it can't close it. Even with a live catalog fetch and correct `webBackendsKnown` handling, an Operator can disable the plugin in the gap between chat.tsx's fetch and the turn actually being sent — `resolveSearchTools` still has to degrade to `{}` for that turn, and something still has to surface that it did. A is additive UX polish on a window that's already rare; B is the part that can't be defeated by a race. That makes B the better first move on more than preference.

## Open questions

1. Which shape do you want — A, B, both, or neither (leave the current warn-log posture as the documented behaviour)?
2. If B: should the notice name the missing backend id? It is free text a chat user cannot act on, but it is what makes a screenshot useful to whoever can.
3. If B: is per-turn message metadata the right channel for "a capability you asked for did not run", and should the key be general rather than search-specific? Leaning general: `ChatMessageMetadata`'s own doc comment (`types.ts:17-24`) already frames it as an additive, optional-key record of per-turn facts, and search is unlikely to be the last backend-runtime concern that silently degrades — a `degradedCapabilities?: string[]` (or similar), with search as its first entry, costs nothing extra to build now and avoids a second one-off boolean the next time this class of gap shows up elsewhere.
4. If A: what should the toggle do while the catalog is unknown — in flight, or after a failed request?
5. Regardless of shape: once a turn's notice has told the user their search didn't run, should `search` flip off client-side for the rest of the session, so they don't keep re-sending into the same dead backend? No server round-trip needed — cheap insurance against repeat waste that doesn't require committing to A.

This comes from #329's deferred follow-ups, where it was recorded as an optional cached liveness check. We are asking for a direction call before writing anything: option B post-dates that follow-up and changes what the right PR looks like.

MesoX:
Closes #522

**Follow-up filed:** #627 — Provider identity on the two `buildTurnTools` warns.

A User turns **Search** on, the Provider names a Web-search backend that is gone or will not start, and the reply is written with no search at all. Until now the only trace was a server-side warn. The reply now carries a notice: **Search was unavailable — this reply was written without it.**

Option **B** from the issue only. The search toggle's visibility is untouched, the model is not told, and no stub Tool is served.

## What changed

**Backend.** Turn resolution derives one fact — search was requested, resolution had somewhere to send it, and no search tools came back — and returns it on `ChatTurn` beside `stream`, not inside `resolved` (which describes the turn's plan and is mirrored into run records). The runner forwards it to the Chat drive on the same path `agentId` takes, and the metadata extractor emits it on the stream's `start` part. `ChatMessageMetadata` gains one optional `true`-literal key.

The check is outcome-based, so one condition covers an unregistered backend id, a `createExecutors` that threw, one that outran its `timeoutMs`, and one that returned no `web_search` executor — plus whatever cause is added next. The three server-side warns keep their distinct messages.

**Frontend.** A new exported notice constant next to `CUT_SHORT_NOTICE`, rendered through the shared `CutShortNotice` row. A reply that lost its search and then hit the output ceiling shows both rows, in that order: how the reply was produced, then how it ended.

**No schema or migration work.** Per-message metadata rides inside the Chat's existing messages JSON column, which is how agent attribution and the cut-short marker already survive a reload.

**Docs, CONTEXT.md and ADR-0014 ship here**, per `CLAUDE.md`. CONTEXT.md defines **Unavailable capability**; ADR-0014's Selection bullet records that the degradation is now reported to the User as well as logged, and that hiding the toggle was rejected rather than deferred.

## Two acceptance criteria worth calling out

**"The notice is present after a page reload."** Satisfied without a dedicated persistence test. The new metadata tests assert against `lastSnapshot`, which consumes the stream the way both the client and the run's own snapshot reader do — that assembled message is exactly what the Chat sink writes to the messages column, and the sink's own tests already cover the write.

**"All three server-side warns ... still carry the Provider id."** Only the registry-miss warn ever did. The other two are raised inside `buildTurnTools`, where `WebBackendContext` is `{ orgId, workspaceId, userId }` and carries no `providerId`; satisfying the criterion literally means widening a plugin-SDK-facing type, which is a separable change. This PR ships the reading that the criterion means "the warns are unchanged, and the one that carries `providerId` keeps it", and adds `orgId`/`workspaceId` to the other two — already in scope, previously unlogged — so an Operator asked about the new notice can at least tell which Workspace saw it. Provider identity is #627.

## Known residual

`searchSource: "native"` on a Provider with no native tool resolves to the none-kind, so a turn there is silently unsearched with no flag and no warn. The UI cannot produce it — the toggle's capability predicate and Turn resolution both gate on the same shared `providerHasNativeSearch` helper — but a direct API caller can. Left as specified, and now documented in ADR-0014.

## Out of scope, per the brief

No toggle hiding and no backend-catalogue fetch in Chat; no System prompt fragment and no stub search Tool; no backend id or cause in the notice; no general capability-degradation channel; no toggle reset after an unavailable-search turn (see #624 for how the toggle's Agent-switch wart interacts with this notice); nothing added to run statistics, the Trigger sink or Trigger run history.

## Verification

`pnpm test` (backend 1947, frontend 458, schemas, plugin-sdk, examples), `pnpm typecheck`, `pnpm lint` — all green. A test pins the boundary the feature rests on: an unavailable-search turn's System prompt and tool-key set are identical to a turn that never asked for search.


willdady:
Reviewed this on two axes (repo standards, and fidelity to #522's Agent Brief) and pushed the remediations directly as 0a13c7e rather than sending it back round — hope that's alright, @MesoX. The design holds up; everything below is either a stated criterion or a naming/structure point. Summary of what changed and why.

## The Provider id on all three warns

This is the one substantive change. #522 asks that "all three server-side warns still fire with their existing distinct messages and still carry the Provider id", and the brief is explicit that the Workspace is not a substitute: "an org-scoped Shared Provider (ADR-0007) is one row serving many Workspaces, so the Workspace alone does not name the row an Operator has to edit."

The PR read this as meaning the warns are unchanged and the one that already carried `providerId` keeps it, on the grounds that satisfying it literally means widening a plugin-SDK-facing type. That reasoning is right about `WebBackendContext` but stops one seam short: `buildTurnTools` is **core's** side of the boundary (`WebBackendRegistration`), not the plugin's — `createExecutors(ctx, plugin)` is the plugin-facing call. So core can widen its own context without touching the published contract:

- `TurnToolsContext = WebBackendContext & { providerId }` in `web-backends/types.ts`, passed to `buildTurnTools`.
- `composeWebBackend` destructures `providerId` off, keeps it for the warns, and forwards the narrow `{ orgId, workspaceId, userId }` to `createExecutors`.

`WebBackendContext` is unchanged, so ADR-0014's **Backend context** consequence stays accurate and no SDK version moves. **#627 is closed as unnecessary** rather than left open.

Worth flagging on the tests here: the previous warn assertions passed while `providerId` was absent, because `toEqual` semantics ignore `undefined` keys — the fixture now carries a real id, so those assertions actually bite. The two `buildTurnTools` warns also get end-to-end coverage in `chat-execution.test.ts`, not just at the compose layer.

## ADR-0014 amendment structure

The two new paragraphs were inserted *inside* the accepted Selection bullet, rewriting decision text in place. This repo labels post-acceptance changes — `## Amendment — \`ValidationError\` → 400 (#478)` (ADR-0010, ×3), `## Amendment (2026-06-13)` (ADR-0011), inline `**Amendment (#312).**` (ADR-0013). Moved to `## Amendment — an unavailable search is reported to the User (#522)`; the Selection bullet is back to its accepted wording. The amendment also records that core's internal context is what widened, so a later reader doesn't mistake it for a breach of the Backend-context decision.

## The reload criterion

"The notice is present after a page reload — i.e. the flag round-trips through the persisted Chat messages." The PR argues this is satisfied without a dedicated test because `lastSnapshot` consumes the stream the way the sink's writer does. The assembled message is right, but the untested step is the one after it: `onFinish` writes messages through `extractFiles`, which rebuilds parts, and nothing asserted it leaves `metadata` alone. Added that assertion to `chat-sink.test.ts`.

## Naming and structure

- **`CutShortNotice` → `TurnNotice`.** The component carries two unrelated notices now, and its doc comment had been reduced to conceding the stale name ("It began as the output-ceiling row and keeps the name"). File renamed, 3 call sites; surface-specific `RunCutShortNotice` untouched.
- **`createMessageMetadata` takes an options object.** `agentId` and `searchUnavailable` are the same kind of setup-time fact and travel together at every hop, so the third positional boolean would have become a fourth. Call sites read `{ agentId: "agent-1", searchUnavailable: true }` instead of `("agent-1", new Map(), true)`.
- **One `faultCtx`** for the two warns in `composeWebBackend`, which had the same four-field literal twice.
- **Dropped "degradation" where it named the concept** — `CONTEXT.md`'s new entry lists it under `_Avoid_`, and the same commit used it in the ADR and a code comment. Code doc comments now reference **Unavailable capability** so the glossary term appears somewhere other than the glossary.

## Left alone

Two low-severity points, called out so they're a decision rather than an oversight:

- The new Callout in `web-search-backends.mdx` partly restates the one above it; the brief asked to extend the existing throwing-factory warning rather than add a second. Cosmetic, and `docs-contract.test.ts` is happy.
- `CONTEXT.md` uses its own prose rather than the definition the brief supplied as quoted text. Substance matches.

The documented residual — `searchSource: "native"` on a Provider with no native tool resolving to the none-kind, unreachable from the UI — is correct as specified and now recorded in the ADR amendment.

`pnpm typecheck`, `pnpm lint`, `pnpm test` all green: backend 1950 (+3), frontend 458.

PR #629, opened 2026-08-21
feat(backend,plugin-sdk): optional per-turn teardown and executor cancellation for search backends and Tool sets

### LINKED ISSUE #523: [Web-search backend] Optional teardown (and cancellation) on the backend and Tool-set Extension points
ADR-0014 defers a teardown hook until a backend needs a real lifecycle, and #513 has since built most of the plumbing that deferral was avoiding — so this asks for direction: take the remaining work now while the seam is fresh, or keep waiting. It is a design question, not a bug report, and the honest position up front is that **the ADR's stated trigger has not been met** — we know of no Web-search backend or Sandbox adapter today holding a browser pool or a keep-alive connection.

## Why this might matter

Two ADR-0014 Consequences point at the same unbuilt piece. (Both were added by #517, the PR that flips 0014 to `accepted`; this issue assumes #516 and #517 have landed, since any work here would follow them.) **"No teardown hook, by design"** names the trigger: _"A backend needing a real lifecycle (a browser pool, a keep-alive connection) is the trigger to add an optional teardown to both Extension points at once, plumbed into the turn's disposal and abort paths with its own timeout, rather than making Web-search backends the only Contribution shape with a lifecycle."_ The timeout bullet defers cancellation to that same change: _"The v1 executor contract also carries no `AbortSignal`: the loser of the race is dropped, not cancelled … Adding cancellation is append-only and belongs with the same PR that adds a teardown hook to both Extension points."_

What changed is the cost, not the trigger. #513 (merged, closing #492) introduced `openToolSession` with a `closers` array (`apps/backend/src/tools/tool-session.ts:100`), an idempotent, never-throwing `dispose` (`ToolSession` at `tool-session.ts:62-76`, implemented `:187-197`) and `nest` for delegates (`:199-213`). The turn opens it at `apps/backend/src/services/chat-execution.ts:564`, hands `session.dispose` to the caller at `:697`, and the caller runs it on abort and on finish alike (`apps/backend/src/runs/agent-runner.ts:179`) — the "turn's disposal and abort paths" the ADR asked for, already built, and reachable by only one Extension point.

## What it would fix

- **A Web-search backend has nowhere to register a closer.** `resolveSearchTools` (`chat-execution.ts:469`) is awaited inside the same `Promise.all` as the session (`:581-586`) but receives only `resolution`, `opened`, `provider` and a `WebBackendContext` — never the session. A backend that opened something in `createExecutors` cannot ask for it to be closed.
- **A cancelled turn does not cancel the upstream call.** `withTimeout` (`apps/backend/src/web-backends/index.ts:156-187`) races and abandons, as its own comment says. Yet the run's `AbortSignal` already reaches tool calls: `agent-runner.ts:239` passes `run.handle.signal` into the model invocation, and the SDK delivers it as `execute`'s second argument, which `apps/backend/src/tools/sub-agent.ts:285-288` destructures. `composeWebBackend`'s executes ignore that argument (`web-backends/index.ts:297`, `:430`), so the signal stops one frame short of the executor contract.
- **An inverted asymmetry.** `composeToolSet` gives Tool-set factories no timeout because _"a ceiling for that belongs on the contribution, where an author can state it"_ (`apps/backend/src/tools/index.ts:160-165`) — precisely what `timeoutMs` already is on a `WebBackendContribution` (`packages/plugin-sdk/index.ts:400`). Tool sets have lifetime management and no budget; Web-search backends have a budget and no lifetime.

## Recommended approach

One PR, three strands, all append-only within API major 1:

1. **Optional teardown on both contracts**, registered into the existing session's `closers` rather than a new mechanism. A `dispose?` on `WebBackendExecutors` (`plugin-sdk/index.ts:362`) is the natural slot; the Tool-set side needs a decision (below). The structural change on the search path should be a narrow `registerCloser(fn): void` passed via `WebBackendContext` into `resolveSearchTools` — not the `ToolSession` itself, which also carries `nest` and internal dispose bookkeeping the plugin-sdk contract has no reason to depend on.
2. **An `AbortSignal` on the executor contract**, forwarded from the signal already arriving at `execute`, so a cancelled turn stops the upstream request instead of abandoning it. This also requires `withTimeout` (`web-backends/index.ts:156-187`) to combine its own timer with the caller's signal (e.g. `AbortSignal.any`) rather than racing them as it does today — a change to `withTimeout`'s shape, not just an additive executor parameter.
3. **Optionally, a per-contribution budget for Tool-set factories**, in the shape #513's "Deliberately not done" section argues for. This is the loosest fit and may be yours to split off; strands 1 and 2 are the ADR's own pairing.

## Open questions

1. **Which two points does "both Extension points" mean?** The bullet's preceding sentence compares against Sandbox adapters, so it reads as Web-search + Sandbox — but #513 has since put Tool sets on the session, which makes Tool sets the more natural pair. Our recommendation: Web-search + Tool-set. Sandbox adapters have `destroy(ctx)` at `plugin-sdk/index.ts:240`, but that is Workspace-lifetime, not per-turn — folding it into the turn's dispose/abort path would mean redefining its lifecycle contract, not reusing one. Tool sets already sit on the same per-turn session `resolveSearchTools` does, so the same registrar mechanism covers both with no second mechanism needed.
2. **How should a Tool set register a closer?** `ToolSetTools` resolves to a plain `Record<string, Tool>` (`plugin-sdk/index.ts:132-137`), so unlike `WebBackendExecutors` there is no returned object to hang a `dispose` on. Our recommendation: a registrar on `ToolSetContext`, mirroring the `registerCloser` shape proposed for the Web-search side (strand 1), rather than a `dispose?` on `ToolSetContribution` — a per-turn registrar keeps both Extension points using the same mechanism and the same disposal granularity (per-turn, not per-contribution), instead of solving the same problem two different ways.
3. **Take it now, or wait?** Now makes both points symmetric in one review while the session is fresh; waiting keeps the surface smaller and honours the ADR's own posture. We have no case that forces it.

This comes from #329's deferred follow-ups, where it was recorded before #513 existed and so assumed far more work than now remains. We are asking for direction rather than proposing to start.

MesoX:
Closes #523.

Strands 1 and 2 of the Agent Brief. Strand 3 — a per-contribution timeout budget for Tool-set factories — is left out, as the brief puts it under Out of scope and core already states the no-timeout position with its rationale.

Two append-only additions to the Plugin contract, both optional, no `PLUGIN_API_VERSION` bump and no new disposal mechanism.

## 1. A per-turn closer registrar

`registerCloser?: (close) => void` on `ToolSetContext` and `WebBackendContext` — the runtime context of the two Extension points whose Contribution is resolved per Chat turn. Registered closers feed the Tool session's existing closer list, so a turn still hands its caller exactly one `dispose`, and it runs on a normal finish and on the User cancelling alike.

The search path is resolved inside the same `Promise.all` as the session, and that concurrency is load-bearing for first-token latency, so the two are not sequenced. The search side gets a registrar that defers onto the session promise; a session that never opens runs the closer itself rather than dropping it.

Optional on purpose. A manifest declares a _major_ `apiVersion` only, so a Plugin cannot require a core new enough to have the member — on an older one it is genuinely absent, and `ctx.registerCloser?.(close)` is the honest shape. Unguarded on old core, the throw costs that turn its tools, with a warn naming the plugin and the Contribution.

Sandbox adapters are deliberately not folded in: `destroy` is Workspace-lifetime, so reusing the per-turn registrar there would redefine that contract rather than reuse this one.

## 2. An `AbortSignal` on the executor contract

`web_search` and `read_url` now receive `{ signal }` as an appended second argument. `withTimeout` becomes `withDeadline`, which _combines_ its deadline with the caller's abort instead of racing a bare timer, so a cooperating backend's upstream request is cancelled rather than merely abandoned — on cancellation and on the deadline alike. It is still a race, so an executor that ignores its signal is bounded exactly as before. A call whose signal has already aborted does not reach the executor at all.

`logCall`'s outcome union gains `cancelled`, at **debug** rather than warn: a User pressing stop is not a fault, and warning on it would make healthy operation noisy. Model-facing strings are unchanged — a cancelled call returns the existing failure message, which in practice the model never reads.

`createExecutors` keeps its current signature. Its own `withDeadline` call passes no caller signal, so factory behaviour is byte-identical; wiring the prepare-phase abort into the factory is a separate, defensible change.

## Worth a reviewer's attention

**Dedupe is per turn.** The identity `Set` lives on the session, so what it collapses is one turn reaching the same teardown twice — two Tool sets from one plugin sharing a client, say. Across turns there is nothing to dedupe against, by design. The consequence is stated on all four surfaces: a module-level `closePool` registered on every turn is closed on the first one, leaving every turn after it holding a closed pool.

**`CLOSER_TIMEOUT_MS = 5_000` is a judgement call, not a derived number.** It exists because the run's `onTerminate` awaits `dispose` before the terminal write, so a hung closer delays the last thing a reader is waiting for. Nothing in the acceptance criteria pins it — a maintainer opinion is welcome. Two limits of the ceiling are recorded in ADR-0014 rather than fixed: it bounds a _closer_, not a _teardown_ (`closers × 5s` on a sequential loop), and a closer registered after `dispose` is fire-and-forget, so teardown _starts_ before the terminal write rather than finishing before it.

**Closer failures log at `warn`, not `error`.** An unreleased connection is worth an Operator's attention and never worth paging for; it cannot fail the turn or reach the User; and most closers that get there belong to third-party code core cannot fix. That is the level the sibling degrade paths already use. `dispose`'s own catch keeps `error`, with a distinct message, because reaching it means an invariant broke rather than a closer merely failing.

**Side effect worth recording:** MCP `client.close()` now runs through the same bounded path, so a hanging MCP close no longer stalls a run's terminal write indefinitely.

## Docs, ADR, CONTEXT.md — in this PR

Every Contribution filling these Extension points is authored outside this repository, so the docs pages are the whole product surface for the change. `extending/web-search-backends.mdx` and `extending/tool-sets.mdx` cover both additions and both hazards (guard the registrar; the signal is optional to read); `extending/plugin-api-and-config.mdx` records the additions against the append-only rule. `docs-contract.test.ts` gains a check pinning `CLOSER_TIMEOUT_MS` against the four surfaces that quote it — an author sizing teardown against a stale figure gets no error, just a close cut off half-done.

ADR-0014's _"No teardown hook, by design"_ is superseded rather than deleted, and — per the issue's closing note — records that its stated trigger was unobservable from a repository that ships no backend of its own, so the decision was taken on the contract's cost to out-of-tree authors instead. The timeout bullet's deferral of cancellation is discharged. CONTEXT.md's **Tool session** claim that a turn has exactly one thing to dispose is now true; the wording is unchanged.

The SDK's `tsconfig` gains `dom` for the one global it now names, `AbortSignal`; `packages/example-plugin` mirrors it, and both docs pages say why an author's own tsconfig may need the same.

## Known adjacent defect — not fixed here

Split out as #630. A Tool session is disposed only through `state.turn`, and there are two windows where it is never assigned in time — `prepareChatTurn` throwing after `openToolSession` resolved, and the run terminating while `prepare` is still in flight. This PR does not create either path, but it widens what they leak: a backend's pool or keep-alive connection as well as MCP clients. Distinct defect, distinct fix, and the acceptance criteria here do not reach it.

## Verification

`pnpm typecheck`, `pnpm lint` and `pnpm --filter backend test` (1953 tests, 126 files) all pass. New tests map to the acceptance criteria across `tool-session.test.ts`, `web-backends/index.test.ts`, `chat-execution.test.ts`, `agent-runner.test.ts` and `plugin-sdk/index.test.ts`, including a pre-change single-argument executor fixture that compiles without `@ts-expect-error`.

One caveat on local runs: `docs-contract.test.ts > internal links` fails on a Windows CRLF checkout, because its heading regex cannot cross a trailing `\r` and so collects no anchors from any `.mdx`. It is unrelated to this change and CI on Linux is unaffected — but it is a real papercut for anyone developing on Windows, and probably worth its own fix.

A manual smoke on the deploy box is still worth doing for the cancellation criterion specifically — cancel a searching turn and confirm the closer's log line — since the abort path is the half unit tests model least faithfully.


willdady:
Rebased onto `main` and pushed directly to this branch, since it had fallen behind `028cea9` — which touched `web-backends/` and `chat-execution.ts` and conflicted with this work in three files. Your two commits are preserved, rebased, authorship intact; their SHAs changed, so reset if you have this checked out.

Full gate green after the rebase: `pnpm typecheck` 7/7, `pnpm lint` 4/4, `pnpm test` 8/8 — 2605 tests. That closes the last acceptance-criteria box, which had been unverified.

## Conflicts, and one real bug in the merge

`028cea9` had introduced `TurnToolsContext = WebBackendContext & { providerId: string }` for core's side of the `buildTurnTools` seam — the same role this branch's `CoreWebBackendContext` filled. Folded into one type:

```ts
export type TurnToolsContext = WithCoreRegistrar<WebBackendContext> & { providerId: string };
```

The consequential one: `028cea9` strips `providerId` before the plugin-facing call, and this branch derived the factory context by spreading core's own. Merged naively, `createExecutors` would have started receiving `providerId` — breaking the invariant `028cea9` states and the `WebBackendContext` shape ADR-0014 fixes. The strip now happens first:

```ts
const { providerId, ...coreCtx } = ctx;
const factoryCtx: WebBackendContext = withAttributedRegistrar(coreCtx, { plugin: pluginName, backend });
```

The registrar-derivation ternary was also duplicated across both Extension points, and that duplication is what let the leak exist on one side only. It is now one helper, `withAttributedRegistrar` in `closers.ts`, shared by both — so the seam is guarded in a single place.

**That path was not pinned.** The exact-match assertion on `createExecutors` only covered the no-registrar case; the registrar-present case — the one where the spread actually happens — used `objectContaining`, which passes whatever else the context carries. Tightened to an exact match, and verified the hard way: reintroducing the leak fails both assertions, restoring it passes.

## Other review follow-ups

**ADR-0014's timeout bullet was rewritten rather than superseded.** It had recorded a decision — cancellation deferred to the PR that adds teardown — and that text was edited away. The v1 position is back under a strikethrough with a supersede note, matching the idiom the teardown bullet beside it already uses. The other three edited bullets I left as they were: `options`, `cancelled` and `registerCloser?` are contract *shape*, not reversed decisions.

**Added the Tool-set half of the version-skew degrade.** Only the Web-search side pinned it, so an unguarded registrar call on an older core was covered by argument rather than by a test.

**The `createExecutors` docs snippet now chains `plugin?.config.endpoint`** — `plugin` is optional in the factory signature, and the SDK README snippet in this same PR already chained.

## One reported finding that did not hold

`withDeadline`'s catch inspecting `deadline.signal.aborted` before `cause` was raised in review as misreporting a genuine executor error as a `timeout`. Probed directly: an early failure reports `error`, a mid-flight cancel reports `cancelled`, and a signal-honouring executor reports `cancelled`. The only `timeout` is when the deadline actually fired, which is what the comment above the function says it intends. Left alone.

## Still open, for you

- **`AbortSignal` forcing `lib: ["dom"]`** on out-of-tree backends is the one acceptance criterion only partially met. It is disclosed in both the ADR and the docs page, and fixing it is a contract call — declare a structural stand-in, or keep requiring the lib — so I did not make it unilaterally.
- Two judgement-call smells left deliberately: the `runCloser`/`withDeadline` race duplication, and `attribution?: Record<string, unknown>`. The current shape is defensible; `attribution` really is "extra log fields", and a union buys rigidity more than clarity.
- `web-search-backends.mdx` is unformatted per Prettier, but it already is on `main`, so reformatting would bloat this diff. Same page, line 272, has a pre-existing unguarded `plugin.config.endpoints` worth its own fix.
- `CLOSER_TIMEOUT_MS = 5_000` is still the judgement call your description flagged. Nothing here pins it; a maintainer opinion is still welcome.

Issue #712, opened 2026-08-27
Switch between Alternatives of edited and regenerated messages


willdady:
Part of #709. Implements the UI half of **ADR-0026** (#988). Blocked by #711.

The earlier body of this issue recorded a proposal and the review that refuted it. The findings that still apply are folded in below, and the issue's history keeps the original text.

## What to build

Let a user see and move between the **Alternatives** that #711 already stores, and regenerate any assistant message.

**PR title:** `feat: keep edited and regenerated messages as Alternatives you can switch between`

### Switch endpoint

- **`PUT /chats/:chatId/active-leaf` with `{ messageId }`:** sets `activeLeafId` to the **newest leaf** under `messageId`, by `createdAt`, skipping deleted rows.
- **Response:** the new `messages` (Active path) and `tree`, so the client swaps in one round trip.
- **Errors:**
  - 409 while the Chat has a run in flight (`status = running`).
  - 404 for an id not in the Chat.
- **Gate:** the same as `POST /chat` (`requireWorkspaceOwner`).

### The `‹ n/m ›` control

- **Where it shows:** on any message on the Active path that has non-deleted siblings, worked out from `tree`.
- **Label by anchor:**
  - Under a user message: "Message n of m".
  - Under an assistant message: "Response n of m".
  - Tooltips: "Previous response", "Next message", and so on.
  - The label doubles as the `aria-label`.
  - Never "version" or "branch" anywhere in the UI.
- **When it's hidden or disabled:**
  - Hidden when `canSendMessages` is false, like the rest of the action bar.
  - Disabled while this tab's turn is `submitted` or `streaming`.
- **Switching:**
  - Optimistic local swap, then the PUT, with **no debounce**.
  - Rapid clicks abort the in-flight request (`AbortController`) instead of delaying the first one.
  - Hold the hydration bypass until the PUT acknowledges. Releasing it on click leaves a window where the 3-second poll lands the old path.
  - On 409 or failure, revert and toast.

### Regenerate on every assistant message

- **Button:** shown on every assistant message on the Active path, not only the last. Remove the `isLastMessage` condition in `chat-message.tsx`.
- **Call:** `regenerate({ messageId, body })`. #711 already made the server side non-destructive.
- **Why no archive step:** the SDK truncates local state from the target onward, which is fine because the server keeps every row. Nothing needs awaiting before the call.

## Acceptance

- [ ] Edit a message, then use the arrows to return to the original: the original's full subtree is shown.
[...]

Issue #711, opened 2026-08-27
Store Chat messages as rows the server owns


willdady:
Part of #709. Implements the storage and ownership half of **ADR-0026** (#988). That PR must merge first.

The earlier body of this issue recorded a `chat.branches` sidecar design and the review that refuted it. That record now lives in ADR-0026's Considered Options. This issue's history keeps the original text.

## What to build

Store every message as a row, make the server the owner of the **Transcript**, and make Delete persist immediately. After this ships, edits and regenerates no longer destroy anything in the database, but there is no UI to reach **Alternatives** yet (#712).

**PR title:** `fix: persist message deletion immediately and stop edits destroying stored messages`

### Schema and migration

- **New `chat_message` table.**
  - Columns: `chatId`, `id`, `parentId` (nullable; null for a message that opens the Chat), `role`, `parts` (jsonb), `metadata` (jsonb), `deletedAt` (nullable), `createdAt`.
  - Primary key `(chatId, id)`.
  - Foreign key to `chat` with `onDelete: cascade`.
  - `(chatId, parentId)` references a row in the same Chat.
- **`chat.activeLeafId`:** nullable.
- **Backfill in `scripts/migrate.ts`, same migration, then drop `chat.messages`.**
  - Each existing array becomes a chain: each message's parent is the message before it, and `activeLeafId` is the last message.
  - Chats with null or empty `messages` get no rows and a null leaf.
- **PR description note:** `drizzle-kit push` does not run the backfill, so dev databases need the migrate script run locally.

### One way to read the Active path

- **`loadActivePath(chatId)`:** walks from `activeLeafId` up through `parentId`, skips rows with `deletedAt` set, and returns `PlatypusUIMessage[]` in order.
- **Every reader goes through it:**
  - `GET /:chatId`
  - `generateChatMetadata` (`services/chat-metadata.ts`)
  - memory extraction (`services/memory-extraction.ts`, `findChatsToProcess` / `formatConversation`)
  - the chat turn's history, which feeds `inlineFileUrls`, `validateTurnAttachments`, `initialOccupancyFrom` and `convertToModelMessages`
- **`findChatsToProcess`** stops loading full rows. It no longer has a `messages` column to load, so it fetches the path per Chat it processes.

### Request contract (`POST /chat`)

- **Submit:** `{ message, parentId, ...existing body }`.
  - `message` must have `role: "user"`, contain only text and file parts, and pass `validateUIMessages`. Otherwise 400.
[...]

Issue #709, opened 2026-08-27
Non-destructive message editing and chat branching


willdady:
Tracking issue for making message editing non-destructive, and for the storage model that makes it possible.

## The problem

Editing a message today drops every message after it, permanently. Every comparable product (ChatGPT, Claude.ai, LibreChat, Open WebUI) keeps the abandoned branch and offers version arrows to return to it.

Separately, the edit surface is a bare `<textarea>` that round-trips text only — so editing a message that has attachments silently discards them today, before branching enters the picture.

## Children

| Ticket | State | Blocked by |
|--------|-------|-----------|
| #710 — Edit through the composer, preserving all message parts | active | — |
| #713 — Skills can be marked user-invocable only | active | — |
| #649 — Slash-command picker in the chat input | active | #710, #713 |
| #711 — Store abandoned branches beside the active path | **parked** | — |
| #712 — Version arrows for edited and regenerated messages | **parked** | — |

#710 and #713 are unblocked and independent of each other.

**Branching (#711, #712) is deferred** — see the comment below. ADR-0022 was refuted in adversarial review and PR #714 is closed; nothing landed on `main`.

#710 is justified on its own regardless: it fixes a live defect where editing a message discards its attachments, with no schema change.

## Acceptance for this tracking issue

Closed when #710, #713 and #649 are closed. #711 and #712 are parked and do not gate this issue; if branching is revived it gets a fresh ADR and its own tracking.

willdady:
**Branching is deferred.** Adversarial review substantially refuted the storage model in ADR-0022; PR #714 is closed and no ADR lands on `main`.

What changes:

- **#710 proceeds** — independently justified by a live defect (editing a message discards its attachments), verified ready, and it stands alone with no schema change.
- **#713 and #649 proceed**, both revised for findings unrelated to branching.
- **#711 and #712 are parked**, not closed. Their bodies carry findings worth keeping; both are off `ready-for-agent`.

If branching is revived, the likely shape is a `message` table with `parentId` plus an `activeLeafId` on `chat` — the option ADR-0022 rejected on cost. The three problems that sank the sidecar model (unguarded whole-column writes from the run sink, a flat marker that cannot express depth, and archived segments riding the 3-second chat poll) all disappear under row-level storage rather than needing individual fixes.

Before reviving it, worth establishing that anyone actually wants version history. The original justification was entirely "every comparable product does this", with no cited user request — a weak basis for knowingly taking on storage debt.

willdady:
[...]

PR #714, opened 2026-08-27
chore: record the chat branching decision

willdady:
Records ADR-0022. No code changes — the decision lands ahead of the work so #711 and #712 can reference it.

## The decision

Editing a message today drops every message after it, permanently. Making that non-destructive is awkward in Platypus because the transcript is not owned by the server: the client sends the entire message array on every turn, and `ChatSink` writes whatever the run produced back over the whole `chat.messages` column on a flush interval.

The ADR records that `chat.messages` **stays exactly as it is** — the linear active path, read by `convertToModelMessages`, memory extraction, auto-titling and context occupancy — and that a sibling `chat.branches` column holds the inactive segments, each tagged with the message it forks from. No downstream reader learns that branching exists.

This is a deliberate compression of the message-tree model that LibreChat and Open WebUI use, not an attempt at it. The ADR says so explicitly, and names the three signals that should trigger migrating to a real message table.

## Why it is worth an ADR

A future reader will look at `branches` sitting beside `messages` and reasonably ask why the transcript is not just a tree. The rejected alternatives are the substance here: a tree inside `chat.messages` is erased by the next sink flush, and storing the active segment inside `branches` creates a second writer for content the sink rewrites continuously.

## Status

`accepted-pending-implementation`, with `implemented-by: "#709"` and an admonition stating what the code does today. Flipping it to `accepted` is an acceptance criterion on #712, the last ticket in the chain.

Closes nothing — #709 tracks the implementation.

### DIFF
--- a/docs/adr/0022-chat-branches-are-archived-segments-beside-the-active-path.md
+++ b/docs/adr/0022-chat-branches-are-archived-segments-beside-the-active-path.md
@@ -1,0 +1,91 @@
+---
+status: accepted-pending-implementation
+implemented-by: "#709"
+---
+
+# Chat branches are archived segments stored beside the active path, not a message tree
+
+> **Not in the code yet.** Today editing a message is destructive: the client
+> truncates the array at the edited message and resubmits, and the dropped
+> messages are gone. `chat.messages` is a single linear jsonb array and there is
+> no `chat.branches` column. The edit surface is also a bare `<textarea>` that
+> round-trips text only, so editing a message with attachments silently drops
+> them — see the first Consequence.
+
+Editing a message in Platypus drops every message after it. Every comparable
+product — ChatGPT, Claude.ai, LibreChat, Open WebUI — keeps the abandoned branch
+and offers version arrows to return to it, and the products that do so store the
+transcript as a tree of message rows with parent pointers. Platypus cannot cheaply
+follow them there, because the transcript is not owned by the server: the client
+sends the entire message array on every turn (`routes/chat.ts`), and `ChatSink`
+writes whatever the run produced back over the whole `chat.messages` column on a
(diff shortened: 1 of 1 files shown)


willdady:
Closing. Adversarial review found three confirmed factual errors in ADR-0022, and the decision it records is substantially refuted:

1. **The `active` marker cannot represent a path through more than one fork point.** After a second edit inside a forked path, every ancestor fork undercounts its siblings by one, and the `createdAt` needed to order them is not recoverable from a single marker. The ADR's Consequence claiming arbitrary-depth support is false — it is a one-level model.
2. **"The sink owns only `messages`" is not an invariant that protects anything.** `ChatSink.onStart` (`apps/backend/src/runs/sinks/chat-sink.ts:66-82`) overwrites the column unconditionally from the client-supplied array. A stale tab submitting after a branch switch permanently loses the segment that switch moved into `messages`. The ADR explicitly claims the sink's contract is untouched; that claim is the data-loss path.
3. **The memory-extraction Consequence describes a mechanism that does not exist.** Nothing resets `memoryExtractionStatus` to `pending` — `updateChatExtractionStatus` (`apps/backend/src/services/memory-extraction.ts:70-79`) is only ever called from inside the extraction job. A Chat is extracted once, ever. Re-extraction after a branch switch cannot happen.

Compounding these, archived segments riding the 3-second chat poll (`CHAT_POLL_INTERVAL_MS`, `apps/frontend/lib/chat-recovery.ts:31`) makes row growth an immediate problem rather than a future migration trigger.

Branching is deferred. If it is built later, a fresh ADR will record the decision — with this refutation as its Considered Options — rather than amending one nobody would trust. The draft remains on `chore/adr-chat-branching` if the reasoning is wanted.

PR #720, opened 2026-08-28
feat: security improvements for path traversal and non-root container

PhoenixSemGarantia:
Hey! 👋

I was reviewing the project and noticed a couple of security improvements that may be worth considering. The project looks great overall, and these changes could help harden the application further.

### 1. Potential path traversal in `disk.ts`

In `disk.ts`, `path.join()` is currently used to construct paths from storage keys.

Depending on how the key value is obtained and validated, a malicious value containing traversal sequences such as `../` could potentially escape the intended storage directory.

A safer approach would be to resolve the final path and explicitly verify that it remains within the configured base directory.

This PR adds a `resolveKey` function that ensures resolved paths cannot point outside the intended storage directory.

### 2. Docker container running as root

I also noticed that the backend Docker image appears to run as root, even though the `hono` user is already created in the Dockerfile.

Running the application as a non-root user reduces the potential impact of a container compromise.

This PR updates the final stage to switch to the existing `hono` user and preserves the correct ownership of copied files using `--chown=hono:nodejs`.

These are relatively small changes, but they provide some additional defense-in-depth.

Great work on the project!

### DIFF
--- a/apps/backend/Dockerfile
+++ b/apps/backend/Dockerfile
@@ -55,11 +55,14 @@ RUN addgroup --system --gid 1001 nodejs && \
     adduser --system --uid 1001 hono && \
     mkdir -p /data/files && chown hono:nodejs /data/files
 
+# Switch to the non-root user
+USER hono
+
 # Copy prod dependencies (includes root node_modules and workspace node_modules)
-COPY --from=prod-deps /app .
+COPY --from=prod-deps --chown=hono:nodejs /app .
 
 # Copy source code
-COPY --from=builder /app/out/full/ .
+COPY --from=builder --chown=hono:nodejs /app/out/full/ .
 
 # Expose port
 EXPOSE 4000
--- a/apps/backend/src/storage/disk.ts
+++ b/apps/backend/src/storage/disk.ts
@@ -17,21 +17,34 @@ export class DiskStorage implements StorageBackend {
   constructor(
     basePath: string = process.env.STORAGE_DISK_PATH || "./data/files",
   ) {
-    this.basePath = basePath;
+    this.basePath = path.resolve(basePath);
+  }
+
+  /**
+   * Safely resolves a key against the base path, preventing directory traversal.
+   */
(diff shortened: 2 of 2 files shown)


willdady:
Thanks for this — the traversal was real, and the report was accurate about both the mechanism and the risk.

It turned out to be worse than the PR described, so we've fixed it at a different layer and landed that as #721 (merged in a0cc599). Two things pushed the fix out of this branch:

**The exposure wasn't mainly the `/files/*` route.** Literal `..` segments are normalized away by the URL parser before routing, and `c.req.path` leaves the rest percent-encoded, so a traversal doesn't survive that path — we checked over a raw socket as well as through the test harness. The reachable hole was `inlineFileUrls`, which reads the key straight out of a client-returned URL string with no parsing at all: a message part carrying `storage://../../etc/passwd` had the file read off the host and inlined into the model prompt.

**Validating inside `DiskStorage` left the other half open.** The same untrusted key reaches `S3Storage` from the same three callers, so the check needed to sit where keys enter the system rather than in one backend. #721 adds `storage/keys.ts` and gates the route, `inlineFileUrls`, and key extraction on it, throwing `ValidationError` so a malformed key answers 400 through the central error seam instead of surfacing as a 500.

While testing we also found a second traversal on the **write** side that this PR didn't reach: `generateStorageKey` interpolates the client-supplied `message.id`, so an id of `../../x` filed uploads outside the Chat and Workspace prefix that `/files/*` parses back to authorize. Also fixed in #721.

**On the container change** — we're not taking it as written, because it stops file storage. `USER hono` lands before the `COPY` instructions, so `WORKDIR /app` stays root-owned and the default `STORAGE_DISK_PATH=./data/files` can't be created; and `compose.yaml` bind-mounts `./data:/data`, which masks the build-time `chown` on Linux. Both give `EACCES` on every upload. The goal is right though, and the frontend image already does it correctly — tracked as #722 with the details.

Closing this in favour of #721 and #722. You're credited in the merged commit. Genuinely useful report — thank you for taking the time.

PR #721, opened 2026-08-28
fix(backend): validate storage keys where they enter Platypus

willdady:
Closes the path traversal reported by @PhoenixSemGarantia in #720, and supersedes that PR's storage half.

## The hole

A **File part**'s storage key round-trips through the client. The read path rewrites `storage://<key>` to an HTTP URL, the **Chat** resubmits its full history every turn, and `inlineFileUrls` reads the key back out of the URL the client returned. Nothing checked it on the way back in.

A part carrying `storage://../../etc/passwd` therefore had its bytes read off the host and inlined into the **System prompt**. Confirmed against a temp root before the fix — the file outside the storage root came back as a `data:` URL on the part.

The `GET /files/*` route turns out **not** to be exploitable this way: `new Request()` normalizes literal `..` segments away before routing, and `c.req.path` leaves the rest percent-encoded, so `..%2f` reaches the backend as a filename rather than a separator. It is still validated here, because that behaviour is the URL parser's and not something this route should depend on.

## The fix

Validation goes where keys **enter** Platypus, not inside one backend — otherwise `S3Storage`, which takes the same untrusted key from the same three callers, stays uncovered.

- **`storage/keys.ts`** states what a key may be: a relative POSIX path whose every segment matches `[A-Za-z0-9._-]`. That admits both shapes Platypus generates and excludes separators, `.`/`..` segments, and percent-encoded forms.
- **`storageKeyFromUrl`** and **`inlineFileUrls`** drop an invalid key, treating it as a URL that named nothing stored. `normalizeFileParts` then announces the part as unavailable, exactly as it already does for a storage miss.
- **`GET /files/*`** throws `ValidationError`, so the seam answers 400 (ADR-0010) rather than letting a filesystem error surface as a 500.
- **`generateStorageKey`** checks the key it composes. `messageId` is the client's, and a message id of `../../x` filed the file outside the Chat and Workspace prefix that `/files/*` parses back to authorize — a second traversal, on the write side, that #720 did not reach.

`DiskStorage` also resolves and re-checks containment as a last line before the filesystem, and derives the `.meta` path from the resolved object path so the object and its sidecar can no longer disagree.

## Not included

The non-root container half of #720 is deliberately left out. As written it moves to `USER hono` before the copies, leaving `WORKDIR /app` root-owned, so the default `STORAGE_DISK_PATH=./data/files` cannot be created and every upload fails with `EACCES`. It needs its own change, with the ownership fixed after the copies and a self-hosting docs note for the `/data` bind mount.

## Tests

`keys.test.ts` covers the key rules directly. Regression tests cover both traversal forms through `inlineFileUrls`, the write-side hostile message id, key-skipping in `extractStorageKeys`, and the route's 400. Backend 2383 passed, frontend 666 passed.

No docs page changes: no `.env.example`, schema limit, plugin, or user-facing label moved.

### DIFF
--- a/apps/backend/src/routes/files.test.ts
+++ b/apps/backend/src/routes/files.test.ts
@@ -46,6 +46,19 @@ describe("Files Routes", () => {
       expect(body.error).toBe("File key required");
     });
 
+    it.each([
+      ["percent-encoded traversal", "/files/..%2f..%2fetc%2fpasswd"],
+      ["percent-encoded dots", "/files/%2e%2e%2fsecret"],
+      ["backslash traversal", "/files/..%5c..%5cwindows%5cwin.ini"],
+    ])("should return 400 for a %s key", async (_label, url) => {
+      mockSession();
+      const res = await app.request(url);
+      expect(res.status).toBe(400);
+      const body = (await res.json()) as Record<string, unknown>;
+      expect(body.error).toBe("Invalid file key");
+      expect(mockStorageGet).not.toHaveBeenCalled();
+    });
+
     it("should return 400 for invalid key format with less than 2 segments", async () => {
       mockSession();
       const res = await app.request("/files/just-one-segment");
--- a/apps/backend/src/routes/files.test.ts
+++ b/apps/backend/src/routes/files.test.ts
@@ -46,10 +46,18 @@ describe("Files Routes", () => {
       expect(body.error).toBe("File key required");
     });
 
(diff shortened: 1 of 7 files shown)


PhoenixSemGarantia:
Thanks! I really appreciate the detailed follow-up and the credit in the merged commit. I'm glad I could help improve the security of the project. I'll keep doing some more analysis when I have time — it's a really interesting project.

PR #727, opened 2026-08-28
fix: bind workspace creation to route organization

PhoenixSemGarantia:
## Summary

Bind new workspaces to the organization authorized by the route. The create payload no longer controls `organizationId`, preventing cross-organization workspace creation.

## Testing

* `pnpm --filter @platypus/backend test -- workspace.test.ts`
* `pnpm --filter @platypus/schemas test -- index.test.ts`
* `pnpm --filter @platypus/backend typecheck`
* `pnpm --filter @platypus/schemas typecheck`

### DIFF
--- a/apps/backend/src/routes/workspace.test.ts
+++ b/apps/backend/src/routes/workspace.test.ts
@@ -42,7 +42,6 @@ describe("Workspace Routes", () => {
         method: "POST",
         body: JSON.stringify({
           name: "New Workspace",
-          organizationId: "org-1",
         }),
         headers: { "Content-Type": "application/json" },
       });
@@ -68,7 +67,6 @@ describe("Workspace Routes", () => {
         method: "POST",
         body: JSON.stringify({
           name: "New Workspace",
-          organizationId: "org-1",
         }),
         headers: { "Content-Type": "application/json" },
       });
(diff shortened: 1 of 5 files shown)


willdady:
Thanks for this, @PhoenixSemGarantia — it was a real cross-tenant bug and the fix was the right shape. It shipped in [v3.1.1](https://github.com/willdady/platypus/releases/tag/v3.1.1) and is published as [GHSA-v34j-qpmw-pxgr](https://github.com/willdady/platypus/security/advisories/GHSA-v34j-qpmw-pxgr).

One ask for next time, plus the reason.

When you opened this, Platypus had no `SECURITY.md`, no issue template and no pull request template. Every documented route into the project pointed at the public tracker, and nothing told you to do anything differently. That gap was ours, not yours.

It's now closed. Private vulnerability reporting is enabled on the repo, and there's a [SECURITY.md](https://github.com/willdady/platypus/blob/main/SECURITY.md) setting out the process. For anything with a security impact, please use the [private report form](https://github.com/willdady/platypus/security/advisories/new) instead of a pull request.

The reason a PR is the wrong shape: every deployment upgrades on its own schedule, so between the PR opening and a release existing, anyone reading the repo learns where the hole is while the affected deployments have nothing to upgrade to. And a PR discloses more than people expect — the branch name, the explanatory comments in the diff, and a test named after the attack each say it as plainly as an issue would.

Reporting privately costs you nothing in credit or authorship. The report itself becomes the draft advisory, and if you already have a patch we can open a temporary private fork and invite you to it, so the fix still lands as your work and you're named when the advisory publishes.

Thanks again. The audit was good work; reviewing it turned up a second finding of the same class, now fixed and published alongside yours.

PR #753, opened 2026-08-29
fix: sort sidebar chats by updatedAt instead of createdAt (#750)

### LINKED ISSUE #750: Sort sidebar chats by updatedAt, not createdAt
### Summary
Sort the sidebar chat list by most recent activity. Today both the backend list query and the sidebar's "Last 7 days" grouping use `createdAt`, so returning to an old chat (or renaming/pinning it) never moves it up the list.

### Steps to reproduce
1. Create chat A.
2. A few days later, create chat B.
3. Send a new message in chat A.

### Expected
Chat A jumps to the top of the sidebar once it has the most recent activity, and re-enters the "Last 7 days" group after being touched.

### Actual
Chat B stays on top: the list is ordered by `createdAt` DESC, and the "Last 7 days" / "Other" split also uses `createdAt`.

### Details
- Backend — `apps/backend/src/routes/chat.ts`, `GET /`:
  ```ts
  .orderBy(desc(chatTable.isPinned), desc(chatTable.createdAt))
  ```
  Change the second key to `desc(chatTable.updatedAt)`.
- Frontend — `apps/frontend/components/app-sidebar.tsx`: the `last7Days` / `other` split filters on `new Date(chat.createdAt)`; switch to `chat.updatedAt` so the grouping matches the ordering.
- `ChatListItem` in `packages/schemas/index.ts` already carries `updatedAt`, and the list endpoint already selects it — no schema or API change needed.

### Notes
One caveat for the implementer: `updatedAt` is also bumped by auto-titling and the memory-extraction job (see the comment in the `chat.ts` POST route), so a chat can move without direct user activity. Sorting on the existing `updatedAt` is the smallest change and matches this request; if that becomes a problem we could introduce an explicit activity timestamp (e.g. `lastTurnAt`) later.

Aditya-233:
## The bug

Today both the backend list query (`GET /api/chat`) and the sidebar's "Last 7 days" grouping use `createdAt`. Returning to an older chat (or renaming/pinning it) never moves it up the sidebar list, even when it has recent activity.

## The fix

- In `apps/backend/src/routes/chat.ts`, sort by `desc(chatTable.updatedAt)` rather than `desc(chatTable.createdAt)`.
- In `apps/frontend/components/app-sidebar.tsx`, compute the `last7Days` / `other` split from `chat.updatedAt` instead of `chat.createdAt`.
- Preserves the primary `isPinned` sort order so pinned chats remain on top.

Closes #750

### DIFF
--- a/apps/backend/src/routes/chat.ts
+++ b/apps/backend/src/routes/chat.ts
@@ -116,7 +116,7 @@ chat.get(
       })
       .from(chatTable)
       .where(whereClause)
-      .orderBy(desc(chatTable.isPinned), desc(chatTable.createdAt))
+      .orderBy(desc(chatTable.isPinned), desc(chatTable.updatedAt))
       .limit(limit)
       .offset(offset);
 
--- a/apps/frontend/components/app-sidebar.tsx
+++ b/apps/frontend/components/app-sidebar.tsx
@@ -151,10 +151,10 @@ export function AppSidebar() {
   const sevenDaysAgo = new Date(now.getTime() - 7 * 24 * 60 * 60 * 1000);
 
   const last7Days = regularChats.filter(
-    (chat) => new Date(chat.createdAt) >= sevenDaysAgo,
+    (chat) => new Date(chat.updatedAt) >= sevenDaysAgo,
   );
   const other = regularChats.filter(
-    (chat) => new Date(chat.createdAt) < sevenDaysAgo,
+    (chat) => new Date(chat.updatedAt) < sevenDaysAgo,
   );
 
   const hasRecent = last7Days.length > 0;


willdady:
Thanks for this @Aditya-233, and sorry for the outcome — the change you wrote is exactly what #750 asked for, but on review the issue itself turned out to be wrong, so I'm closing this while I rethink the approach. That's on me for labelling it `good first issue`; it isn't one.

The problem is that `updatedAt` isn't an activity signal on `chat`. Three writers bump it with no user involvement:

- `services/chat-metadata.ts` (auto-titling) — harmless in practice, it only fires just after the first turn when the Chat is top of the list anyway.
- `services/memory-extraction.ts` — mostly harmless for the same reason, but the `failed` branch retries hourly, so a Chat whose extraction keeps failing would be bumped every hour and stick to the top of the sidebar indefinitely.
- `services/model-alias-migration.ts` — a bulk update that stamps one identical `updatedAt` across every matching Chat in the Workspace, floating them all up and collapsing their relative order into ties.

`schema.ts` actually documents this on the `lastTurnAt` column: ADR-0020 measures idle gaps against `lastTurnAt`, "never against `updatedAt`", for precisely this reason.

So the right column is `lastTurnAt`, which is written only by the run sink at turn boundaries. That's more than a one-line swap though: it has no backfill (nullable, NULL for every Chat older than migration `0053`), so it needs `COALESCE(lastTurnAt, createdAt)`; and it's deliberately stripped from the Chat response, so the sidebar's "Last 7 days" grouping would need a derived field on `ChatListItem` — which contradicts the "no schema or API change needed" line in the issue.

There's also a separate gap either way: the sidebar list is only revalidated on rename/delete and first title generation, so sending a message won't reorder it live regardless of the sort key.

I've left #750 open with the details and removed the `good first issue` / `ready-for-agent` labels until I've settled the design. If you'd like to pick it up once that's decided, you'd be very welcome — and if you'd rather take a different issue in the meantime, happy to point you at one.

PR #754, opened 2026-08-29
feat(frontend): animate context meter entrance

### LINKED ISSUE #751: Animate the occupancy meter's row height when it first appears on mobile
### Summary

Follow-up polish for the Context occupancy meter (#449). On mobile, when the first assistant message of a chat returns, the meter appears instantly in its own row, which jumps the composer's height. The row should animate from 0 to its natural height over roughly a second when it appears.

### Steps to reproduce

1. Make sure the chat's selected model has a declared Context window in its settings — the occupancy gauge only shows when this is declared.
2. On a mobile viewport, open a chat with no assistant messages yet.
3. Send a message and let the first assistant message come back.
4. Watch the composer footer — the occupancy meter row appears instantly, snapping the composer's height up.

### Expected

The occupancy meter row transitions in smoothly, animating its height from 0 (plus a fade-in, if it reads better) over about a second, so the composer grows gently instead of jumping.

### Actual

The meter row appears immediately at full height, snapping the composer taller with no transition.

### Environment

- Mobile frontend (`apps/frontend`), composer footer
- Related: #449 — the meter itself; this is purely the entrance-animation behaviour on first appearance

### Notes

- Only the first-appearance transition is in scope here — no change to subsequent re-renders.
- Height animation should respect `prefers-reduced-motion` if that's already handled elsewhere in the app.

ump45nose:
## What

- animate the mobile context-meter row only when occupancy becomes available after the initial render
- use the existing Motion animation stack with `AnimatePresence initial={false}`
- keep desktop and reduced-motion layouts immediate
- preserve the desktop toolbar width and prevent the collapsed mobile row from pulling the toolbar upward
- cover initial presence, later mobile appearance, rerender stability, desktop, and reduced-motion behavior

## Why

The meter can become available after an assistant response supplies occupancy data. Its first mid-chat appearance should be smooth on mobile, while a meter already present during initial render must remain stable.

Closes #751.

## Checklist

- [x] The **PR title** is a Conventional Commit subject.
- [x] **The change is covered by tests.**
- [x] No documentation update is needed because no visible label, field, limit, enum, or configuration changed.
- [x] `pnpm test` passes: 8/8 Turbo tasks (schemas 155, docs 49, backend 2,462, frontend 707, plugin SDK 15, example tool set 5).
- [x] `pnpm typecheck` passes: 8/8 Turbo tasks.
- [x] `pnpm lint` passes: 4/4 Turbo tasks.
- [x] `pnpm build` passes: 5/5 Turbo tasks.
- [x] Prettier and `git diff --check` pass.

Local verification used Node 22.23.2, so pnpm printed the Node >=24 engine warning; all commands above completed successfully. CI will provide the supported Node 24 run.

AI assistance was used to implement and verify this change; I reviewed the diff and take responsibility for it.

### DIFF
--- a/apps/frontend/components/chat.tsx
+++ b/apps/frontend/components/chat.tsx
@@ -13,6 +13,7 @@ import {
 import { useChat } from "@ai-sdk/react";
 import { DefaultChatTransport } from "ai";
 import { GlobeIcon, Info, Settings2 } from "lucide-react";
+import { AnimatePresence } from "motion/react";
 import { useRef, useEffect, useState, useCallback, useMemo } from "react";
 import {
   Chat as ChatType,
@@ -42,7 +43,7 @@ import { useSearchToggle } from "@/hooks/use-search-toggle";
 import { useModelSelection } from "@/hooks/use-model-selection";
 import { resolveModel } from "@/lib/resolve-model";
 import { clearedToolCallIds } from "@/lib/tool-result-clearing";
-import { ContextMeter } from "./context-meter";
+import { ContextMeter, ContextMeterEntrance } from "./context-meter";
 import { useMessageEditing } from "@/hooks/use-message-editing";
 import { ATTACHMENTS_ONLY_TEXT } from "@/lib/message-parts";
 import { useChatTitlePoll } from "@/hooks/use-chat-title-poll";
@@ -726,11 +727,21 @@ export const Chat = ({
                   // space, which reads as the last item of the tool row rather
                   // than drifting into the middle the way `justify-between`
                   // alone would leave it.
-                  <ContextMeter
-                    className="order-last -mx-3 -mb-3 mt-1.5 w-[calc(100%+1.5rem)] justify-center rounded-b-md bg-foreground/5 px-3 py-1.5 sm:order-none sm:mx-0 sm:mt-0 sm:mb-0 sm:w-auto sm:justify-start sm:rounded-none sm:bg-transparent sm:p-0 sm:mr-auto"
-                    occupancy={projectedOccupancy}
-                    contextWindow={resolvedModel?.contextWindow}
-                  />
+                  <AnimatePresence initial={false}>
+                    {projectedOccupancy != null &&
+                      resolvedModel?.contextWindow != null && (
+                        <ContextMeterEntrance
+                          key="context-meter"
+                          className="order-last -mx-3 w-[calc(100%+1.5rem)] sm:order-none sm:mx-0 sm:mr-auto sm:w-auto"
+                        >
+                          <ContextMeter
+                            className="mt-1.5 justify-center rounded-b-md bg-foreground/5 px-3 py-1.5 sm:mt-0 sm:justify-start sm:rounded-none sm:bg-transparent sm:p-0"
+                            occupancy={projectedOccupancy}
+                            contextWindow={resolvedModel.contextWindow}
+                          />
+                        </ContextMeterEntrance>
+                      )}
+                  </AnimatePresence>
                 }
                 submit={<PromptInputSubmit status={effectiveStatus} />}
               />
--- a/apps/frontend/components/context-meter.test.tsx
+++ b/apps/frontend/components/context-meter.test.tsx
@@ -1,7 +1,112 @@
-import { describe, it, expect } from "vitest";
-import { render, screen } from "@testing-library/react";
+import { afterEach, describe, it, expect, vi } from "vitest";
+import { render, screen, waitFor } from "@testing-library/react";
+import { AnimatePresence } from "motion/react";
 
-import { ContextMeter } from "./context-meter";
+import { ContextMeter, ContextMeterEntrance } from "./context-meter";
+
+const motionPreference = vi.hoisted(() => ({
+  isMobile: true,
+  reducedMotion: false,
+}));
+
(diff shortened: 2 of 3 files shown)


willdady (dismissed):
Thanks for this — the entrance reads well and the tests are a genuine attempt at pinning it. I'd like it reworked to use `motion` before it lands, and there are a few findings worth folding in at the same time.

## Please use `motion` rather than a hand-rolled reveal

`motion` is already a dependency (`^12.43.0`) and is how every other animated component in the app does this — `expandable-textarea.tsx`, `pull-to-refresh.tsx`, `kanban-board.tsx`, `tag-input.tsx`, `ai-elements/shimmer.tsx`. The `requestAnimationFrame` + `grid-rows-[0fr]→[1fr]` trick works, but it's a second animation idiom in the codebase and it's the reason for the `data-context-meter-entrance` test hook and the class-string assertions.

It also fixes the correctness problem below for free: `<AnimatePresence initial={false}>` animates only children that *appear after* the first render, which is exactly the scope the issue asks for.

Roughly:

```tsx
// chat.tsx
<AnimatePresence initial={false}>
  {projectedOccupancy != null && resolvedModel?.contextWindow != null && (
    <ContextMeterEntrance key="context-meter">
      <ContextMeter ... />
    </ContextMeterEntrance>
  )}
</AnimatePresence>
```

with the wrapper animating `height: 0 → "auto"` and `opacity`, using `useReducedMotion()` from `motion/react` and `useIsMobile()` (`hooks/use-mobile.ts`) to collapse the duration to 0 for reduced-motion and desktop.

## The animation is keyed to mount, not to first appearance

> Only the first-appearance transition is in scope here — no change to subsequent re-renders.

`entered` starts `false` on every mount, so opening a chat that *already has* assistant messages plays the full 1s grow-in on mobile — every page load, and every navigation between chats. The issue's scenario is the meter becoming available mid-chat; this also animates when it was never absent. `AnimatePresence initial={false}` is the fix.

## Desktop layout regression

Pre-change the flex item in the footer row was `ContextMeter`, which carries `shrink-0` in its base class. Now the flex item is the wrapper (`sm:block sm:w-auto sm:mr-auto`), which has no `shrink-0`, and its child is `min-h-0 overflow-hidden`. On a tight desktop toolbar the wrapper can shrink below content width and the child clips the `42% · 84K/200K` label instead of holding its width. Worth carrying `shrink-0` onto whatever ends up being the flex item.

## Smaller things

- **The null guard now lives in two places.** `chat.tsx` gained `projectedOccupancy != null && resolvedModel?.contextWindow != null ?`, which re-implements `ContextMeter`'s own `if (occupancy == null || contextWindow == null) return null`. The guard genuinely needs to move up for the animation to work, so this is justified — but `ContextMeter`'s doc comment still claims the job ("Renders nothing unless BOTH numbers are known") and now overstates. Either update the comment or drop the internal guard and let the type require both.
- **`-mb-3` stays on the wrapper while it's collapsed at `0fr`**, so for the frame before `entered` flips there's a 12px negative margin with no row to offset it — a small upward pop before the grow.
- **`duration-1000`** is 2× the slowest transition anywhere in `components/` (7 × `duration-200`, one each of 300/500). The issue does say "about a second", so this is defensible — just flagging it as an outlier if it feels slow in the hand.
- **`prefers-reduced-motion` is new to the codebase.** Grepping `motion-reduce` / `prefers-reduced-motion` / `useReducedMotion` across `apps/frontend` returns nothing outside this diff, so the issue's "if that's already handled elsewhere" doesn't apply — this is the first instance. Handling it is the right call; `useReducedMotion()` makes it the same idiom as the rest of the animation.

## Tests

The new test asserts Tailwind class strings (`toHaveClass("duration-1000", "motion-reduce:transition-none", "sm:block")`), which can't verify what the test name claims — jsdom applies no `sm:` or `motion-reduce:` CSS, so "without animating desktop or reduced-motion layouts" is untested; only class presence is. Neither the desktop clipping above nor the reduced-motion path is actually covered. Every other test in the file asserts rendered text and ARIA roles.

Moving the decisions into `useIsMobile()` / `useReducedMotion()` makes them mockable, so the tests can assert behaviour — "renders at natural height immediately when reduced motion is on", "does not animate a meter present at first render" — instead of class names, and the `data-context-meter-entrance` attribute can come out of the production DOM.

ump45nose:
Implemented the requested rework and force-pushed the rebased branch:

- replaced the requestAnimationFrame/CSS-grid reveal with `motion`
- wrapped the conditional meter in `AnimatePresence initial={false}`, so a meter present on initial render does not animate
- used `useIsMobile()` and `useReducedMotion()`; desktop and reduced-motion layouts render immediately
- added `shrink-0` to the animated flex item
- moved the mobile negative margin inside the clipped content, avoiding the collapsed-state toolbar jump
- updated the ContextMeter documentation around the caller-level null gate
- removed the production test hook and replaced class-string assertions with behavior tests

Validation on the updated branch:
- `pnpm test` — 8/8 tasks passed (schemas 155, docs 49, backend 2,462, frontend 707, plugin SDK 15, example tool set 5)
- `pnpm typecheck` — 8/8 tasks passed
- `pnpm build` — 5/5 tasks passed
- Prettier and `git diff --check` passed

The local environment uses Node.js 22 while the repository requires Node.js 24+, so pnpm emitted an engine warning; all commands completed successfully.

Issue #549, opened 2026-08-18
Invitations carry a token and a single-use link that redeems into a new account


willdady:
> *This was generated by AI during triage.*

Design settled in #462 with @MesoX. This is the first of two shipping units; the second (`REQUIRE_INVITATION_TO_SIGN_UP`) depends on this one and is filed separately. The decision record is ADR-0019, and the domain terms **Invitation** and **Invitation link** are defined in `CONTEXT.md`.

Related: #548 fixes mixed-case invitation addresses. The two touch the same row and the same handlers; whichever lands second should expect a small conflict.

## Agent Brief

**Category:** enhancement
**Summary:** Give an Invitation a token and a single-use link that redeems it into a new account, joining the Organization and provisioning the Workspace in one act

**Current behavior:**
An Invitation is a row scoped to an Organization, matched to a person by email address. To act on one, the invitee must **already hold an account with that exact address** — the invitation only becomes visible once such an account exists, through the header notification or the in-app Invitations list. So onboarding is two disjoint steps in a fixed order: the person self-registers, and only then can they see and accept what was sent to them. There is nothing the sender can hand over that does anything: no link, no token, no artifact. The sender's only option is to tell the person out of band to go and register, then look for a notification.

Accepting is an authenticated action that joins the Organization as `member` and provisions a Workspace, applying the Invitation's ordered Blueprint set (ADR-0008, ADR-0009).

**Desired behavior:**
An Invitation carries a token minted when it is created, and there is a redemption route keyed on that token — `/invite/<token>` — that is the entrance for a person who does not yet have an account.

Opening a valid redemption link, unauthenticated, presents a registration form that already knows who the person is: the email address comes off the Invitation and is not editable. They supply their name and choose their own password. On submission, in one act: the account is created, they are signed in, and the Invitation is accepted with everything acceptance already does — Organization membership, Workspace provisioning, Blueprints applied in order.

The route serves all three arrival states:

- **No account** — as above.
- **Signed in as the invited address** — no registration is needed or offered; the Invitation is accepted and they land in the new Workspace.
- **Signed in as a different account** — refused, stating plainly that the link was issued for a different address and offering to sign out. It must never bind the Invitation to the signed-in account.

A pending Invitation's link is obtainable from the Organization's Invitations surface as a copy action on the row, available for as long as the Invitation is pending — not only at the moment of creation.

The existing in-app acceptance path is unchanged and remains the route for someone who already holds an account, or who has lost the link.

**Key interfaces:**

[...]

Issue #550, opened 2026-08-18
REQUIRE_INVITATION_TO_SIGN_UP: let an Operator require an invitation to register


willdady:
> *This was generated by AI during triage.*

Design settled in #462 with @MesoX. Second of two shipping units. **Blocked by #549** — this closes public registration, and #549 builds the entrance that has to exist before the door can be shut. The decision record is ADR-0019.

## Agent Brief

**Category:** enhancement
**Summary:** Let an Operator require an Invitation to register, closing public sign-up without removing self-service registration

**Current behavior:**
Self-service registration is open and nothing Platypus exposes turns it off. Anyone who can reach the frontend can create an account: the sign-in page links to a sign-up form, and the account exists on submission with no address verification. The underlying auth library has an option to disable it, but Platypus never sets it, so closing registration today means patching and rebuilding the backend.

The mitigating half is that a self-registered account has no Organization and no Workspace — but it is still an account row, and it is enough to be invited into an Organization later.

Once #549 lands, an Invitation can be redeemed into an account through its link, which means there is finally an entrance that does not depend on public registration being open.

**Desired behavior:**
An Operator can require that an account only comes into existence by redeeming an Invitation. When required:

- The public registration endpoint is **closed**, using the auth library's own supported option, not a middleware or request filter that inspects registration attempts. There is no request that creates an account except through the Invitation redemption route.
- Sign-in is entirely unaffected, as are password changes and administrative password resets.
- Invitation redemption works exactly as #549 built it. Registration is still **self-service**: the invitee chooses their own name and password, and no initial password is ever set by, transmitted through, or known to an administrator. Only the entrance changes.
- The frontend does not present a dead end: the **Sign up** link on the sign-in page is not shown, and the registration page itself, if reached directly, states that registration is by invitation and does not present a submittable form.

The requirement **defaults to off**, so an upgrading deployment's admission posture is unchanged until an Operator opts in. The default is expected to flip at the next major version.

**Key interfaces:**

- **`REQUIRE_INVITATION_TO_SIGN_UP`** — a backend environment variable, boolean, default off. Named for what it governs, at registration; it must not read as governing Organization membership or Workspace access, which it does not touch. Parsed the same way the codebase's other boolean environment variables are.

[...]

PR #850, opened 2026-09-15
feat: complete invitation link redemption

NewPixel1987:
Completes #549.

An invitation link now handles all three ways someone can arrive:

- Signed out: the page resolves the token and shows the invited address and
  the Organization name. The address is fixed, so they only pick a name and
  a password.
- Signed in as the invited address: a single Accept button.
- Signed in as someone else: the page explains the mismatch and offers
  sign-out. The Invitation stays pending.

Redemption creates the account through the auth library's admin create-user
API, signs the person in, then accepts through the same accept path the
in-app flow uses. Either way they land in the Workspace that was just
provisioned. Once the page has resolved the token it's removed from the
address bar, and a bare /invite shows a short "open your link again" page
for reloads.

The Invitations table has a copy-link action for as long as an invitation is
pending, and the token backfill is limited to pending, unexpired invitations.

Two things beyond the spec:

- The signed-out form has an "Already have an account? Sign in" toggle. It
  covers the 409 case, where an account already exists for the invited
  address, so they can sign in and accept without leaving the page.
- The in-app accept endpoint now returns the organization and workspace ids
  too, since it shares a response helper with the link routes.

Docs (self-hosting/users-and-access) and ADR-0019 are updated in this PR.

Validation:

- backend invitation tests
- frontend invite-page tests
- local Postgres check for the invitation migrations and accept flow
- backend, frontend, and docs typechecks


willdady (changes requested):
Thanks for this — the security thinking is the right thinking. The identical-response guard across unknown/non-pending/expired is exactly what #549 asked for, and the test that asserts the three responses are indistinguishable is the test I'd have asked you to write.

I can't merge it as scoped, though. Anything merged to `main` goes out in the next release, so a PR has to fully implement its linked issue — a partial slice ships a `token` column and a public endpoint that nothing in the product uses yet, and that's a release note for a feature nobody can reach. Splitting is genuinely reasonable engineering, it just doesn't fit how this repo releases.

To close out #549, this needs the rest of it:

- **`/invite/<token>`** serving all three arrival states — no account (registration form with the address fixed and non-editable, plus the Org name), signed in as the invited address (accept, no registration offered), signed in as a different address (refuse, explain, offer sign-out, leave the Invitation `pending`).
- **Server-side redemption** — account creation via the auth library's administrative user-creation API rather than the public registration endpoint, then sign-in, then acceptance. Acceptance must call the existing accept path, not a second provisioning fork.
- **Copy-link action** on the Organization's Invitations surface, available the whole time the Invitation is pending.
- **Token cleared from the visible URL** once the page has resolved it.
- **Tests** for redemption by a new account, redemption while signed in as the invited address, refusal as another account, and an already-redeemed token. The expired case you already have.
- **Docs** — `apps/docs/content/building-with-platypus/`. They ship in the same PR as the code here, never as a follow-up.

Two smaller notes on what's here:

1. The backfill mints a token for every row, including `accepted`, `declined` and expired ones. #549 scopes it to pending and unexpired. The resolution endpoint rejects them anyway so nothing leaks, but it's a wider write than specified.
2. `gen_random_uuid()::text` gives backfilled rows a 36-char token against 21 for freshly minted ones, so the column carries two formats permanently. The entropy reasoning in your comment is sound — this is about there being one shape. Worth considering whether the backfill can run through `scripts/migrate.ts` in JS with the app's generator instead.

Also heads up: #548 touches the same row and handlers, so expect a small conflict depending on which lands first.

willdady (changes requested):
Thanks for this — the backend shape is right: single shared accept path, oracle-safe resolution endpoint, and the `FOR UPDATE` race fix is sound. A short round of changes before merge.

### Requested changes

- **Land in the new Workspace after redemption.** #549: "Signed in as the invited address — … they land in the new Workspace." Both success paths do `router.push("/")` (`apps/frontend/app/invite/[token]/page.tsx:82,122`), which resolves to the first org's home list — for a user already in other orgs that may not even be the invited org. The accept path already knows the provisioned workspace; return its id (and org id) from `/register` and `/accept` and push to it.
- **Docs: state the password minimum.** `packages/schemas/index.ts` adds `password.min(8)`, but `apps/docs/content/self-hosting/users-and-access.mdx` only says "their name and a password". Repo rule: a user-facing `min` in `packages/schemas` updates the matching docs page in the same PR. Give the number.
- **ADR-0019 is now stale.** Header still reads `status: accepted-pending-implementation` / `implemented-by: "#549, #550"`, and the banner says "None of this is built yet… #549 builds the token". Append a short note that #549 is implemented by this PR (don't rewrite the body); precedent is `6bb09b81`.
- **Stale comment** in `apps/backend/src/db/schema.ts:557-560` — "resolve — and, in a later change, redeem". This PR is that change.

### Suggestions (non-blocking)

- Run Prettier: `apps/backend/src/routes/invitation-link.ts:169`, `invitation-link.test.ts:151,230,247` reflow.
- Loading labels use `…`; neighbouring `sign-in`/`sign-up` pages use ASCII `...`.
- Tests: expired and already-redeemed tokens are only exercised on `GET /:token`. Worth a case each on `POST /register` and `POST /accept` so the acceptance-criteria line is covered directly rather than via the shared resolver.
- The outcome→response mapping (`not_found`→404, `expired`→410, else `{message}`) appears three times (`invitation-link.ts:165-172, 205-212`, `user-invitation.ts:58-66`). One shared mapper or the typed-error seam would state it once.
- After the URL scrub to `/invite`, a reload 404s since no page exists there. Either a stub page that says the link needs re-opening, or a note in the PR that it's accepted.
- `mockCreateUser` default arg in `test-utils.ts:365` is never used.

### Noted, fine as-is

- Inline "Already have an account? Sign in" toggle is a fourth arrival state #549 didn't define. It's sensible for the 409 case — please mention it explicitly in the PR body so the squashed history is honest about it.
- Backfill uses `gen_random_uuid()::text` rather than the nanoid shape. The migration comment justifies it; no objection.
- Commit `a54b87d9` lacks a Conventional Commit type, but the squash uses the PR title, which is correct.

NewPixel1987:
Thanks for the detailed review, and for bearing with the back and forth.
Pushed a follow-up commit.

Requested changes: /register and /accept now return the org and workspace
ids and the page pushes to /{org}/workspace/{ws}. The docs give the
8 character minimum, and the stale comment in schema.ts is gone.

On ADR-0019 I edited the banner rather than appending a note, since "none of
this is built yet" would otherwise sit right above a note saying the
opposite. It stays accepted-pending-implementation with implemented-by
"#550", because the Operator switch isn't in yet. Easy to change if you'd
prefer a different shape.

Suggestions: I took all of them. Prettier, ASCII ellipses, expired and
redeemed cases on register and accept (plus the case where the invite goes
away between resolving the token and accepting), one shared response helper,
a small /invite page so a reload doesn't 404, and the unused mockCreateUser
default is gone. The sign-in toggle is now in the PR body.

One side effect: /users/me/invitations/:id/accept now returns the ids too.
The alternative was leaving that route alone, which would have kept a second
copy of the outcome-to-response mapping you suggested merging. Nothing in the
repo reads that response body (the Settings page only checks that it
succeeded), so it shouldn't affect anything, but I wanted to flag it.

On the token handling I'm comfortable with where this landed. A used,
declined or expired token is refused before /register creates an account,
and each case has a test on both routes. I checked that those tests fail
if the pending/expiry check is removed. Acceptance is single-use because the
status flips inside the same FOR UPDATE transaction, and an unknown, spent or
expired token gets the same 404 body.

The part I'm less sure of is the better-auth side. /register calls
auth.api.createUser with no headers, relying on it being treated as a
trusted internal call, and then relays the session cookie from signInEmail.
I checked the admin plugin source in 1.6.25 and a headerless call is
accepted, but I'm depending on library internals, and the route tests mock
better-auth, so nothing exercises the path for real. If you want to try it,
create a fresh invitation first, since ones from before the token column
existed have no token (push skips the backfill). If you know of a reason
that's fragile, please let me know.

willdady (commented):
Standards pass over the diff (`main...` head). Branch prefix, commit types, and the PR title all conform — no issues there.

## Needs fixing

**1. `apps/backend/drizzle/meta/_journal.json` — trailing newline stripped**

The renumber commit dropped the trailing newline. `prettier --check` passes on `main`'s copy of this file and fails on this branch's, and it's the only changed file that fails. `.prettierignore` doesn't cover `drizzle/`, so this will show up in the format gate. One-line fix.

**2. `apps/backend/src/routes/invitation-link.ts` — no scoping middleware named**

Our `add-api-endpoint` guidance ends with "every route in the module names its scoping middleware explicitly." These routes mount at a top-level `/invitation-links` with none. The token *is* the credential here and the JSDoc makes that case well, so this reads as a deliberate and correct exception rather than an oversight — but nothing at the module or mount level states it. Worth one explicit comment so the next reader doesn't have to reconstruct the reasoning.

## Docs

No breach of the docs table. The new `password: z.string().min(8)` is user-facing, but invitations are documented in `self-hosting/users-and-access.mdx` — which this PR updates, and which now states the 8-character minimum. Two smaller things:

- The new "They open the link" section mixes third and second person inside one list ("What they see depends on whether they're signed in", then "Enter your existing password… You stay on the page."). House style is second person throughout.
- The docs bold **Copy invitation link** as though it were a visible label; in `page.tsx` it's an icon-only button whose only text is a `title` tooltip.

## Judgement calls

Take or leave these — none of them block.

- **`invitation-link.ts` register handler** detects "account already exists" with `e.body?.message?.toLowerCase().includes("already exists")`. That's reading better-auth's message text; a wording change on their side silently turns a 409 into a 500. Worth keying off something more stable.
- **`acceptResultResponse`** puts Hono `Context` and status mapping inside `services/`, so that module now changes for both domain and transport reasons. Our typed-error seam covers `NotFoundError` → 404; the `{ outcome: "expired" }` → 410 case has no equivalent, which seems to be the only thing holding it there.
- **`app/invite/[token]/page.tsx`** repeats the `flex min-h-screen items-center justify-center` + `max-w-md … text-center` wrapper five times across its four states plus `app/invite/page.tsx`.
- **`const invalid = !isLoading && (error !== undefined || data === null)`** followed by `if (invalid || !data)` — the guard restates the flag, so the name promises more than it delivers.

The `fake-db.ts` / `test-utils.ts` stubs and the `.for("update")` lock look right, and the schema change ships with its `.sql`, snapshot, and journal entry as required.

A separate pass on this against #549 turned up some things too; I'll follow up on those separately.

NewPixel1987:
Thanks for the standards pass, and for the detail. Pushed a follow-up.

Both fixes are in. `_journal.json` has its trailing newline back, and `invitation-link.ts` now says at the router why these routes have no scoping middleware: the token is the credential, and accept adds `requireAuth` on its own.

On the docs, the sign-in bullet now describes the invitee in third person like the rest of the section, and the copy action is described as the icon and its tooltip instead of a bold label.

I went with all four judgement calls. The duplicate-account check now reads better-auth's error code (`USER_ALREADY_EXISTS_USE_ANOTHER_EMAIL`) instead of the message text. `acceptResultResponse` moved from `services/` to `routes/`; I didn't try to fit the 410 into the typed-error seam. The invite page states share one wrapper, and the `invalid` flag is gone. I confirmed the error code by reading the admin plugin in 1.6.25, but the test only covers it with a mock, so a real duplicate registration is the thing I'm least sure of.

I also merged main, since it had rewritten `AuthProvider` and moved `useBackendUrl` into it. The branch now only adds `refreshSession` there.

I left the 0069 and 0070 snapshots as generated output, since main's own 0066 and 0067 don't pass Prettier either. Easy to reformat if you'd rather.

willdady (approved):
Both items from the standards pass are addressed on the current head: the `_journal.json` trailing newline is back, and `invitation-link.ts` now documents why it mounts without scoping middleware.

